PR #40, opened 2021-11-19
Implement k8s unit tests with mocks; implement get api and url methods

holisticode:
This PR adds mocking ability to the complete kubernetes with Avalanchego operator stack.


k8s/node.go
```diff
@@ -56,3 +59,13 @@ func (n *Node) GetURI() string {
 func (n *Node) GetK8sObject() *k8sapi.Avalanchego {
 	return n.k8sObj
 }
+
+// GetStakingURL
+func (n *Node) GetStakingURL() string {
+	return fmt.Sprintf("%s:%d", n.k8sObj.Status.NetworkMembersURI[0], constants.DefaultStakingPort)
```
danlaine, line 65: Can we just use `n.uri` instead of `n.k8sObj.Status.NetworkMembersURI[0]`?

k8s/node.go
```diff
@@ -56,3 +59,13 @@ func (n *Node) GetURI() string {
...
+}
+
+// GetAPIURL
+func (n *Node) GetAPIURL() string {
+	return fmt.Sprintf("%s:%d", n.k8sObj.Status.NetworkMembersURI[0], constants.DefaultAPIPort)
```
danlaine, line 70: Can we just use `n.uri `instead of `n.k8sObj.Status.NetworkMembersURI[0]`?

k8s/network.go
```diff
@@ -198,7 +211,7 @@ func (a *networkImpl) AddNode(cfg node.Config) (node.Node, error) {
 	}
 
 	uri := node.Status.NetworkMembersURI[0]
-	cli := api.NewAPIClient(uri, constants.DefaultPort, constants.APITimeoutDuration)
+	cli := api.NewAPIClient(uri, constants.DefaultAPIPort, constants.APITimeoutDuration)
```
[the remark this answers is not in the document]
holisticode: Well that's what the Avalanchego operator sets. I think we can assume this the same way as with a common avago node. If you don't provide any env var, then the default will be taken.

We could therefore check if the user provided a custom port, but to be honest, it would not make much sense for the user to do that - it's a pod, not a process, therefore pods don't share their port number space...

k8s/node.go
holisticode: See above

k8s/node.go
holisticode: We could, although the `Status` object holds the actual response from k8s. `n.uri` _should_ contain the same....

constants/constants.go
```diff
@@ -5,6 +5,8 @@ import "time"
 const (
 	HealthCheckInterval = 3 * time.Second
 	APITimeoutDuration  = 10 * time.Second
-	DefaultPort         = 9650
+	DefaultAPIPort      = 9650
+	DefaultStakingPort  = 9651
```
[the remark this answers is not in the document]
holisticode: Yeah every node gets its own port, so as long as we as users don't provide fancy custom ports, it can be fairly reasonably assumed that the operator will set the default

constants/constants.go
```diff
@@ -5,6 +5,8 @@ import "time"
...
 	DefaultNetworkID    = uint32(1337)
```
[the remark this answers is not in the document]
holisticode: created package `constants/testconstants`...

constants/constants.go
```diff
@@ -5,6 +5,8 @@ import "time"
...
+	DefaultNetworkSize  = 5
```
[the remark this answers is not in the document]
holisticode: `constants/testconstants`...

constants/constants.go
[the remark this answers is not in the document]
danlaine: I don't think we should hardcode the ports like this. The user could give a different API port in a node's config file, and then we won't be able to make API calls to a given node since we assume they're listening on 9650, right?

k8s/http.go
```diff
@@ -0,0 +1,21 @@
+package k8s
+
+import "net/http"
+
+type dnsCheck interface {
```
danlaine, line 5: Specification please

k8s/http.go
```diff
@@ -0,0 +1,21 @@
...
+	Reachable(url string) error
+}
+
+type DefaultDNSChecker struct{}
```
danlaine, line 9: We don't have to export this

k8s/http.go
```diff
@@ -0,0 +1,21 @@
...
+func NewDefaultDNSChecker() *DefaultDNSChecker {
+	return &DefaultDNSChecker{}
+}
+
+func (d *DefaultDNSChecker) Reachable(url string) error {
+	var err error
+	if _, err = http.Get(url); err == nil {
+		return nil
+	}
+	return err
```
danlaine, line 20:
```suggestion
	_, err := http.Get(url)
	return err
```

k8s/network.go
```diff
@@ -48,49 +46,72 @@ const (
...
+	config network.Config
+	k8scli k8scli.Client
 	// Node name --> The node
 	nodes     map[string]*Node
 	beaconURL string
 	// Closed when network is done shutting down
 	closedOnStopCh chan struct{}
 	log            logging.Logger
+	dnsChecker     dnsCheck
+	apiClientFunc  api.NewAPIClientF
```
danlaine, line 71: comments please

k8s/network.go
```diff
@@ -48,49 +46,72 @@ const (
...
-		nodes:          make(map[string]*Node, len(conf.NodeConfigs)),
+		log:            params.log,
+		nodes:          make(map[string]*Node, len(params.conf.NodeConfigs)),
+		dnsChecker:     params.dnsChecker,
+		apiClientFunc:  params.apiClientFunc,
 	}
+	net.log.Debug("launching beacon nodes...")
 	// Start the beacon nodes
+	defer func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
danlaine, line 109: Let's make `10*time.Second` a constant called `stopTimeout` or similar

k8s/node.go
danlaine: I think it'd be cleaner to use `n.uri`. If it ever happens that `n.uri` !=  `n.k8sObj.Status.NetworkMembersURI[0]` then something is wrong and we'd want to know

danlaine (changes requested):
Didn't have time to look at `network_test.go` but left comments elsewhere

constants/constants.go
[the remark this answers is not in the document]
holisticode: Really the moment we use kubernetes the user should not think about ports at all. Every node gets its own pod and there is simply no reason to change the port...

Having said that, the other question is if the operator is actually passing any custom ports to the node and actually assigning such a port. It might just be assigning the defaults and and not allowing any others for networking reasons. We'll have to ask @yevhenvolchenko about this.

k8s/http.go
holisticode: Agreed, although this results in this linter comment from the `revive` linter:
```
 unexported-return: exported func NewDefaultDNSChecker returns unexported type *k8s.defaultDNSChecker, which can be annoying to use (revive)
```

constants/constants.go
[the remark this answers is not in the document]
yevhenvolchenko: operator will ignore both `DefaultAPIPort` and `DefaultStakingPort`
ref: https://github.com/ava-labs/avalanchego-operator/blob/main/controllers/avalanchego_controller.go#L90

actual values indeed are 9650 and 9651
ref: https://github.com/ava-labs/avalanchego-operator/blob/main/controllers/avalanchego_controller_obj.go#L317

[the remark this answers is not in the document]
danlaine: @holisticode OK, if the operator ignores these ports and uses hard-coded ones then let's:
1) Export the hard-coded values (9650 and 9651) from the operator
2) Import them in the K8s runner

constants/testconstants/testconstants.go
```diff
@@ -0,0 +1,11 @@
+package testconstants
+
+import "time"
+
+const (
+	DefaultNetworkID   = uint32(1337)
+	DefaultNetworkSize = 5
```
danlaine, line 7: These should be in a `_test.go` file since they're only used in tests. 

```
StopTimeout        = 10 * time.Second
RemoveTimeout      = 30 * time.Second
PollTimeout        = 5 * time.Second
```
^ I'd prefer these go in the `k8s` package since they're only used there. Seems reasonable to use different timeouts for local runner and k8s runner

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/ethereum/go-ethereum v1.10.12
 	github.com/fatih/color v1.10.0 // indirect
 	github.com/gorilla/handlers v1.5.1 // indirect
+	github.com/gorilla/mux v1.8.0
```
danlaine, line 13: This isn't used -- you can do `go mod tidy` to remove this

danlaine (approved):

PR #42, opened 2021-11-19
don't export DefaultNetworkConfig and don't overwrite its fields

danlaine:


local/network.go
```diff
@@ -206,34 +211,45 @@ func newNetwork(
...
 	binaryPath string,
 	newAPIClientF api.NewAPIClientF,
 	newNodeProcessF NewNodeProcessF,
 ) (network.Network, error) {
-	for i := 0; i < len(DefaultNetworkConfig.NodeConfigs); i++ {
-		DefaultNetworkConfig.NodeConfigs[i].ImplSpecificConfig = NodeConfig{
+	config := defaultNetworkConfig
+	// Don't overwrite [DefaultNetworkConfig.NodeConfigs]
+	config.NodeConfigs = make([]node.Config, len(defaultNetworkConfig.NodeConfigs))
+	copy(config.NodeConfigs, defaultNetworkConfig.NodeConfigs)
```
felipemadero, line 245: something is redundant? either the copy func or the element copy inside the for iteration. am i missing something?

danlaine: you're right, removed the redundant line

felipemadero (approved):

PR #46, opened 2021-11-29
os.Exit(1) upon error; remove newClientFunc

danlaine:


examples/k8s/main.go
```diff
@@ -49,27 +55,28 @@ func main() {
...
-		os.Exit(1)
+		return err
 	}
 	networkConfig.ImplSpecificConfig = allConfig.K8sConfig
 
 	level, err := logging.ToLevel(networkConfig.LogLevel)
 	if err != nil {
-		log.Warn("Invalid log level configured: %s", err)
+		log.Fatal("couldn't parse log: %s", err)
+		return err
```
holisticode, line 79: I was not exiting at this point - I understood this to be such a low-impact error that I decided to simply ignore this.

k8s/network.go
```diff
@@ -120,10 +111,14 @@ func newNetwork(params networkParams) (network.Network, error) {
 
 // NewNetwork returns a new network whose initial state is specified in the config
 func NewNetwork(conf network.Config, log logging.Logger) (network.Network, error) {
+	k8sClient, err := newK8sClient()
+	if err != nil {
+		return nil, fmt.Errorf("couldn't create k8s client: %w", err)
```
holisticode, line 116: Didn't we say we use `%s` for error messages formatting

examples/k8s/main.go
danlaine: If `err != nil` than `level` isn't well-defined, and this should never happen, so we should exit here

k8s/network.go
danlaine: Inside of `log.Info`, `log.Debug`, etc., yes, because those methods don't work with the `%w` directive.

In `fmt.Errorf` it doesn't _really_ matter if we do `%` or `%s` for errors. `%w` "wraps" the error but we never use that functionality here: https://go.dev/blog/go1.13-errors

holisticode (approved):

PR #48, opened 2021-12-01
Change k8s namespace; clean Healthy() call handling

danlaine:


holisticode (approved):
Please note that now local runs will fail until a local namespace `ci-avalanchego` exists.

We should probably also bump the avalanchego version?
And we should think of a way to possibly automate this.

danlaine:
`And we should think of a way to possibly automate this`

I don't think we want to automatically update the AvalancheGo dependency, in case there's a breaking change, for example.

PR #61, opened 2021-12-13
add method GetAllNodes to network.Network interface; honor method inv…

danlaine:
…ariants about returning ErrStopped after Stop() in k8s runner

### DIFF
--- a/k8s/network.go
+++ b/k8s/network.go
@@ -154,6 +154,10 @@ func (a *networkImpl) GetNodeNames() ([]string, error) {
 	a.nodesLock.RLock()
 	defer a.nodesLock.RUnlock()
 
+	if a.isStopped() {
+		return nil, network.ErrStopped
+	}
+
 	nodes := make([]string, len(a.nodes))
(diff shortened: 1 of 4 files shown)


k8s/network.go
```diff
@@ -232,6 +244,20 @@ func (a *networkImpl) Stop(ctx context.Context) error {
...
+	a.nodesLock.RLock()
+	isStopped := a.isStopped()
+	a.nodesLock.RUnlock()
+
+	// TODO fix this is race condition:
+	// Stop() can be called after the lock is released above.
+	// If that happens, we'll create the pod inside launchNodes
+	// and then never destroy it.
+	// launchNodes assumes [a.nodesLock] isn't held so we can't just
+	// hold the lock inside this method to fix this race condition.
```
holisticode, line 256: `launchNodes` could have a `select` with a `closedOnStop` case?

danlaine: Fabio and I talked about this and we'll patch in a follow-up PR

PR #64, opened 2021-12-20
examples/local/fivenodenetwork: add logging, set public-ip to skip NAT traversal

gyuho:
Add color logging for local processes

Screen Shot 2021-12-19 at 9 39 39 PM
Screen Shot 2021-12-19 at 9 39 50 PM

### DIFF
--- a/examples/local/fivenodenetwork/main.go
+++ b/examples/local/fivenodenetwork/main.go
@@ -71,18 +71,11 @@ func run(log logging.Logger, binaryPath string) error {
 	config := local.NewDefaultConfig(binaryPath)
 	config.LogLevel = "ERROR"
 	for i := range config.NodeConfigs {
-		config.NodeConfigs[i].ConfigFile = []byte(`{
-	"network-peer-list-gossip-frequency":"250ms",
-	"network-max-reconnect-delay":"1s",
-	"public-ip":"127.0.0.1",
-	"health-check-frequency":"2s",
-	"api-admin-enabled":true,
-	"api-ipcs-enabled":true,
-	"index-enabled":true
-}`)
+		nodeName := fmt.Sprintf("node-%d", i)
+		config.NodeConfigs[i].Name = nodeName
 		wr := &writer{
 			col:  colors[i%len(config.NodeConfigs)],
-			name: fmt.Sprintf("node%02d", i+1),
+			name: nodeName,
 			w:    os.Stdout,
 		}
 		config.NodeConfigs[i].ImplSpecificConfig = local.NodeConfig{
--- a/examples/local/fivenodenetwork/main.go
+++ b/examples/local/fivenodenetwork/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"os/signal"
 	"syscall"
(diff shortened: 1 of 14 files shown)


examples/local/fivenodenetwork/go.mod
```diff
@@ -0,0 +1,109 @@
+module github.com/ava-labs/avalanche-network-runner/examples/local/fivenodenetwork
```
danlaine, line 1: What's does having another `go.mod` / `go.sum` do here?

examples/local/fivenodenetwork/main.go
```diff
@@ -99,3 +118,23 @@ func run(log logging.Logger, binaryPath string) error {
...
+var colors = []*color.Color{
+	color.New(color.FgGreen),
+	color.New(color.FgYellow),
+	color.New(color.FgBlue),
+	color.New(color.FgMagenta),
+	color.New(color.FgCyan),
+}
+
+type writer struct {
+	colIdx int
```
danlaine, line 131: Should we have `col *color.Color` here instead so that we don't need to index into `colors` each time?

examples/local/fivenodenetwork/main.go
```diff
@@ -99,3 +118,23 @@ func run(log logging.Logger, binaryPath string) error {
...
+	name   string
+	f      *os.File
```
danlaine, line 133: Should we make this `w io.Writer` to make this more general?

examples/local/fivenodenetwork/main.go
```diff
@@ -66,7 +67,25 @@ func main() {
...
+	"health-check-frequency":"2s",
+	"api-admin-enabled":true,
+	"api-ipcs-enabled":true,
+	"index-enabled":true
+}`)
+		wr := &writer{colIdx: i, name: fmt.Sprintf("node%02d", i+1), f: os.Stdout}
+		config.NodeConfigs[i].ImplSpecificConfig = local.NodeConfig{
+			BinaryPath: binaryPath,
+			Stdout:     wr,
+			Stderr:     wr,
```
danlaine, line 85: If we provide `Stdout` and `Stderr`, it'll print all the logs from each node to the terminal, which is kind of messy. I'm OK with this change if we change the log display level to `ERROR` in each node config (and keep the log level written to disk at the default, `INFO`.)

danlaine (changes requested):

examples/local/fivenodenetwork/go.mod
gyuho: Trying not to bring `github.com/faith/color` dependency to the main repo. If we are open to it, I can remove this workaround :)

danlaine: Ah, I see. Makes sense

examples/local/fivenodenetwork/main.go
```diff
@@ -66,7 +66,16 @@ func main() {
...
+	config := local.NewDefaultConfig(binaryPath)
+	config.LogLevel = "ERROR"
+	for i := range config.NodeConfigs {
+		nodeName := fmt.Sprintf("node-%d", i)
+		config.NodeConfigs[i].Name = nodeName
+		config.NodeConfigs[i].ImplSpecificConfig = local.NodeConfig{
+			BinaryPath: binaryPath,
+		}
+	}
+	nw, err := local.NewNetwork(log, config)
```
danlaine, line 78: We can remove these changes -- all this is doing is changing the log level of the network runner (as opposed to any of the nodes running in the network) to `ERROR`, which I don't think we want to do

danlaine (approved):

PR #66, opened 2021-12-26
local: move "Stdout/err" to "node.Config"

gyuho:
https://github.com/ava-labs/avalanche-network-runner/pull/58 changed `ImplSpecificConfig` to `[]byte`, so `io.Writer` with JSON tag doesn't make sense anymore. Move to the top level config.


danlaine:
I'm not sure that it makes sense to put `Stdout` and `Stderr` as fields in `node.Config` because only the local implementation of the network runner uses those fields

danlaine:
@gyuho 
There isn't a notion of redirecting stdout and stderr on a node running on a k8s instance so I made [this PR](https://github.com/ava-labs/avalanche-network-runner/pull/70), which fixes the issue (that you can't unmarshal an `io.Writer`) but keeps that config in the local node config rather than the top level node config

PR #67, opened 2021-12-29
CHS 260 integrate ava-sim

holisticode:
This PR integrates the https://github.com/ava-labs/ava-sim repo into the network runner and allows to run custom VM binaries with the network runner


danlaine:
When this is merged, also merge https://github.com/ava-labs/avalanche-docs/pull/488/files

vms/subnet.go
```diff
@@ -0,0 +1,257 @@
...
+		[]string{fundedAddress},
+		fundedAddress,
+		[]string{fundedAddress},
+		defaultKeyThreshold,
+	)
+	if err != nil {
+		return fmt.Errorf("unable to create subnet: %w", err)
+	}
+
+CREATE_SUBNET:
```
felipemadero, line 99: not very fan of labels in control flow. probably status get and check may be done outside of select .

vms/subnet.go
```diff
@@ -0,0 +1,257 @@
...
+			if status.Status == platformvm.Committed {
+				break CREATE_SUBNET
+			}
+			log.Debug("waiting for subnet creation tx (%s) to be accepted", subnetIDTx)
+		}
+	}
+	log.Info("subnet creation tx (%s) accepted", subnetIDTx)
+
+	// Confirm created subnet appears in subnet list
+	subnets, err := txPChainClient.GetSubnets([]ids.ID{})
```
felipemadero, line 115: Cant just query for the specific subnet?

vms/subnet.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot query subnets: %w", err)
+	}
+
+	var rSubnetID ids.ID
+	var subnetID string
+	for _, s := range subnets {
```
felipemadero, line 122: this is not so clear, I think the main goal was to check is it is found. rSubnetID and subnetID both depend on vm.SubnetID. All of this can be improved with a flag and those vars obtained from vm.SubnetID.
Anyway will be better to pass specific ID to txPChainClient.GetSubnets

vms/subnet.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+
+	ADD_VALIDATOR:
+		for {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(loopTimeout):
+				status, _ := txPChainClient.GetTxStatus(txID, true)
+				if status.Status == platformvm.Committed {
```
felipemadero, line 153: Again, can get status outside of select and avoid label. Also, it should have a helper in this file for this kind of tx wait.

vms/subnet.go
```diff
@@ -0,0 +1,257 @@
...
+		return fmt.Errorf("could not read genesis file (%s): %w", vm.Genesis, err)
+	}
+	txID, err := txPChainClient.CreateBlockchain(
+		userPass, []string{fundedAddress}, fundedAddress, rSubnetID,
+		vm.ID, []string{}, vm.Name, genesis,
+	)
+	if err != nil {
+		return fmt.Errorf("could not create blockchain: %w", err)
+	}
+STATUS:
```
felipemadero, line 174: same as before

vms/subnet.go
```diff
@@ -0,0 +1,257 @@
...
+	if blockchainID == (ids.ID{}) {
+		return errors.New("could not find blockchain")
+	}
+
+	statusCheckTimeout := longTimeout
+	// Ensure all nodes are validating subnet
+	for _, n := range allNodes {
+		nodeID := n.GetNodeID().PrefixedString(constants.NodeIDPrefix)
+		nClient := n.GetAPIClient().PChainAPI()
+	VALIDATING:
```
felipemadero, line 210: same comment here. can avoid label.

vms/subnet.go
```diff
@@ -0,0 +1,257 @@
...
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(statusCheckTimeout):
+				status, err := nClient.GetBlockchainStatus(blockchainID.String())
+				if err != nil {
+					return fmt.Errorf("error querying blockchain status: %w", err)
+				}
+				if status == platformvm.Validating {
+					// after the first acceptance, next nodes probably don't need to check that long anymore
+					statusCheckTimeout = loopTimeout
```
felipemadero, line 222: probably do this outside of the loop?

vms/subnet.go
```diff
@@ -0,0 +1,257 @@
...
+			}
+		}
+		log.Info("%s validating blockchain %s", nodeID, blockchainID)
+	}
+
+	// Ensure network bootstrapped
+	for _, n := range allNodes {
+		nodeID := n.GetNodeID().PrefixedString(constants.NodeIDPrefix)
+		nClient := n.GetAPIClient().InfoAPI()
+	BOOTSTRAPPED:
```
felipemadero, line 235: same comment

felipemadero (commented):
nice functionality

vms/subnet.go
```diff
@@ -0,0 +1,257 @@
...
+	validatorStartDiff = 30 * time.Second
+	validatorEndDiff   = 30 * 24 * time.Hour // 30 days
+)
+
+// SetupSubnet creates the necessary transactions to create a subnet for a given custom VM.
+// It then also waits until all these transactions are confirmed on all nodes.
+// Finally it creates a blockchain and makes sure all nodes of the given network
+// are validating this blockchain.
+// It requires a `privateKey` in order to issue the necessary transactions
+func SetupSubnet(
```
felipemadero, line 43: should this be part of the network API?

felipemadero (commented):
Also dont know if this functionality should be integrated in the network interface

programs/local/customvm/main.go
```diff
@@ -0,0 +1,138 @@
...
+		if err := vms.SetupSubnet(
+			subnetCtx,
+			log,
+			v,
+			nw,
+			local.DefaultNetworkFundedPrivateKey); err != nil {
+			return err
+		}
+	}
+	<-utils.WatchShutdownSignals(log, nw.Stop)
```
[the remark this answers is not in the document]
holisticode: Great catch, thanks

vms/subnet.go
holisticode: It could. I thought it to have it very separate first.

vms/subnet.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return err
+	}
+
+	txNodeNames, err := network.GetNodeNames()
+	if err != nil {
+		return err
+	}
+	// txNode will be the node we issue all transactions on
+	txNode := allNodes[txNodeNames[0]]
```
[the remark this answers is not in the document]
holisticode: `allNodes` is a map

vms/subnet.go
holisticode: but we have to check for `ctx.Done()` ?

holisticode: I could extract these to individual funcs - let's wait for @danlaine 's review.

vms/subnet.go
holisticode: Yep solved by the above comment

vms/subnet.go
felipemadero: I mean this (think should work)

```
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-time.After(loopTimeout):
		}
		status, _ := txPChainClient.GetTxStatus(subnetIDTx, true)
		if status.Status == platformvm.Committed {
			break
		}
		log.Debug("waiting for subnet creation tx (%s) to be accepted", subnetIDTx)
	}
	log.Info("subnet creation tx (%s) accepted", subnetIDTx)
```

vms/subnet.go
```diff
@@ -0,0 +1,323 @@
...
+			}
+			if status == platformvm.Validating {
+				// after the first acceptance, next nodes probably don't need to check that long anymore
+				statusCheckTimeout = loopTimeout
+				break
+			}
+			args.log.Debug("waiting for validating status for %s", nodeID)
+		}
+		args.log.Info("%s validating blockchain %s", nodeID, blockchainID)
+	}
```
danlaine, line 298: In all these places that we loop over all nodes, and then have an inner loop that includes listening for a context to finish, we should use an error group. The inner for loop here will be the function run by each goroutine in the error group. If one of them returns an error (due to its context being canceled, for example) then all of them should have their context canceled.

programs/local/customvm/main.go
```diff
@@ -0,0 +1,165 @@
...
+	subnetIDKey    = "subnet-id"
+	vmIDKey        = "vm-id"
+)
+
+var (
+	goPath            = os.ExpandEnv("$GOPATH")
+	defaultBinaryPath = fmt.Sprintf("%s%s", goPath, "/src/github.com/ava-labs/avalanchego/build/avalanchego")
+)
+
+type customVMConfig struct {
```
danlaine, line 33: This is actually multiple custom VM configs, not one, isn't it?
Would it make more sense to have this be
```
type customVMConfig struct {
	VMPath     string
	VMGenesis  string
	VMSubnet   string
	VMID       string
}
```
or similar? And then pass around `[]customVMConfig` where we now pass around just `customVMConfig`

danlaine (changes requested):

vms/subnet.go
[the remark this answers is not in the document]
felipemadero: We have this problem in helper function on testing framework (and probably in tests) that can receive an empty network.

local/network.go
```diff
@@ -264,6 +268,75 @@ func NewDefaultConfig(binaryPath string) network.Config {
...
+	if len(customVms) == 0 {
+		return network.Config{}, fmt.Errorf("No custom vm information has been provided - a default network should be created if this is intended")
+	}
+	for _, v := range customVms {
+		if _, err := os.Stat(v.Path); os.IsNotExist(err) {
+			return network.Config{}, fmt.Errorf("%s vm path does not exist", v.Path)
+		}
+		if _, err := os.Stat(v.Genesis); os.IsNotExist(err) {
+			return network.Config{}, fmt.Errorf("%s genesis does not exist", v.Genesis)
+		}
```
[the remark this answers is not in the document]
holisticode: This is true, nevertheless there are so many steps in between until `createBlockchain` that I think it's justified to not read it until then?

local/network.go
```diff
@@ -264,6 +268,75 @@ func NewDefaultConfig(binaryPath string) network.Config {
...
+			return network.Config{}, fmt.Errorf("the provided subnetID %s failed validation", v.SubnetID)
+		}
+		vmBinDir := filepath.Dir(v.Path)
+		binDir := filepath.Dir(binaryPath)
+		if binDir != vmBinDir {
+			pluginDir := filepath.Join(binDir, "plugins", v.Name)
+			if err := utils.CopyFile(v.Path, pluginDir); err != nil {
+				return network.Config{}, fmt.Errorf("failed to copy binary %s to pluginsDir %s: %w", v.Path, pluginDir, err)
+			}
+		}
```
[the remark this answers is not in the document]
holisticode: Basically this compares the provided vm binary path directory with the avalanchego plugins binary path directory. If they are the same, we assume the user already copied it there, so we don't do nothing.

But there was a bug so thanks for the comment :D

In its basics this is reused from the ava-sim repo.

local/network.go
```diff
@@ -264,6 +268,69 @@ func NewDefaultConfig(binaryPath string) network.Config {
...
+	for i := 0; i < len(config.NodeConfigs); i++ {
+		// TODO can we use just binary path or do we need a customvm jsonraw
+		config.NodeConfigs[i].ImplSpecificConfig = utils.NewLocalNodeConfigJsonRaw(binaryPath)
+		config.NodeConfigs[i].ConfigFile = extendConfigFile(config.NodeConfigs[i], customVms)
+	}
+	return config, nil
+}
+
+func extendConfigFile(config node.Config, customVms []vms.CustomVM) string {
+	var ws strings.Builder
```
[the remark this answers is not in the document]
holisticode: The PR for this is pending. I will thus just remove this code for now, and will have to re-test when the PR is merged.

programs/local/customvm/main.go
holisticode: Of course it's possible, and also makes sense. But I don't think we gain much here. This is just the main for an example/program. We later in `run` create actual individual instances of `vms.CustomVM`, and pass this `customVMConfig` "around" just once to `run`.

I'd say we could keep it as-is.

programs/local/customvm/main.go
```diff
@@ -0,0 +1,165 @@
...
+	log, err := logFactory.Make("main")
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	v := viper.New()
+	v.SetDefault(binaryPathKey, defaultBinaryPath)
+
+	bp := pflag.String(binaryPathKey, defaultBinaryPath, "Path to avalanchego binary")
```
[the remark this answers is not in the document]
holisticode: I think my idea initially here was that I would get a more flexible config with Viper for "free" (config file, env vars, cli vars) - e.g. kinda define-once-and-get-it-all. I guess that is not the case.

So the alternative here is to just remove Viper and just use pflags. I think this is justified as, again, this is just an example and we really just want to read from the command line - unless we want to "institutionalize" this example. In this case, we should define what we want from this (cli-flag support? env vars? file? all of it?) and take a decision.

programs/local/customvm/main.go
```diff
@@ -0,0 +1,165 @@
...
+
+	watchShutdown := utils.WatchShutdownSignals(log, nw.Stop)
+
+	healthyChan := nw.Healthy(ctx)
+	log.Info("waiting for all nodes to report healthy...")
+	if err := <-healthyChan; err != nil {
+		return err
+	}
+	// use a new timed context as we need to wait for the validators validation start time
+	subnetCtx, subnetCancel := context.WithTimeout(ctx, healthyTimeout)
```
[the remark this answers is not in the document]
holisticode: I thought I wanted to use that timeout's _length_, not instance really.
I think it makes sense to define a new one with same length and assign here.

utils/shutdown.go
```diff
@@ -0,0 +1,45 @@
...
+	closedOnShutdownCh := make(chan struct{})
+	go func() {
+		shutdownOnSignal(log, shutDownFunc, signalsChan, closedOnShutdownCh)
+	}()
+
+	return closedOnShutdownCh
+}
+
+// Blocks until a signal is received on [signalChan], upon which
+// [n.Stop()] is called. If [signalChan] is closed, does nothing.
```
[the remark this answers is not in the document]
holisticode: I wanted to first see if this approach would be accepted before changing all occurrences.

vms/subnet.go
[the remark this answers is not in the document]
holisticode: If we have an empty node names array then we have a serious problem imo. Currently solving by first checking if the len is 0. 

I think we should hide this from the user by either having `GetFirstNode` and/or `GetRandomNode` in the framework.

vms/subnet.go
```diff
@@ -0,0 +1,323 @@
...
+				break
+			}
+			args.log.Debug("waiting for subnet creation tx (%s) on (%s) to be accepted", subnetIDTx, s)
+		}
+	}
+	args.log.Info("all nodex accepted subnet tx creation")
+	return nil
+}
+
+func isSubnetInList(args *args) error {
```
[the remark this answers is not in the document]
holisticode: OK. Although in this case no values are changed, and with passing just one arg we can cover all args and possible even if something might change in the future.

vms/subnet.go
```diff
@@ -0,0 +1,323 @@
...
+	// confirm created subnet appears in subnet list
+	_, err := args.txPChainClient.GetSubnets([]ids.ID{args.rSubnetID})
+	if err != nil {
+		return fmt.Errorf("subnet not found: %w", err)
+	}
+	return nil
+}
+
+func addAllAsValidators(args *args, vm CustomVM) error {
```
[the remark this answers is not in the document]
holisticode: OK. `subnetID string` though.

vms/subnet.go
```diff
@@ -0,0 +1,323 @@
...
+	var blockchainID ids.ID
+	for _, blockchain := range blockchains {
+		if blockchain.SubnetID == args.rSubnetID {
+			blockchainID = blockchain.ID
+			break
+		}
+	}
+	if blockchainID == (ids.ID{}) {
+		return errors.New("could not find blockchain")
+	}
```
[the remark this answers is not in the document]
holisticode: Mindless copy paste from original. Thanks.

vms/subnet.go
```diff
@@ -0,0 +1,323 @@
...
+	// Print endpoints where VM is accessible
+	args.log.Info("Custom VM endpoints now accessible at:")
+	for _, n := range args.allNodes {
+		args.log.Info("%s: %s:%d/ext/bc/%s", n.GetNodeID(), n.GetURL(), n.GetAPIPort(), blockchainID.String())
+	}
+	return nil
+}
+
+func ensureValidating(args *args, blockchainID ids.ID) error {
```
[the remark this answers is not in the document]
holisticode: OK. 

TBH I pretty dislike this style...That's the whole point of the `args` object , to contain all variables needed for the functions. Also, I see it as pretty elegant that most functions end up having a very similar interface and short argument lists. It helps for readability a lot IMHO.

And finally, with this style, when there is a refactor, you need to touch a lot of code, made similar changes all over the place, change all function signature and possibly tests.

While with a uniform data holder object, it's usually just adding/removing the member in question.

programs/local/customvm/main.go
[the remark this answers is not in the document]
danlaine: Let's use viper and support CLI flags, env vars and config file

programs/local/customvm/main.go
```diff
@@ -0,0 +1,199 @@
...
+	VMIDs        []string `json:"vmIds"`
+}
+
+// Shows example usage of the Avalanche Network Runner with custom VMs.
+// Creates a local five node Avalanche network all starting an avalanchego binary
+// and a provided custom VM binary. Waits for all nodes to become healthy.
+// The network runs until the user provides a SIGINT or SIGTERM.
+// Example of how to run this:
+// go run programs/local/customvm/main.go --vm-path "/path/to/vm/binary" --genesis-path "/path/to/genesis/file" --subnet-ids "24tZhrm8j8GCJRE9PomW8FaeqbgGS4UAQjJnqqn8pq5NwYSYV1" --vm-ids "tGas3T58KzdjLHhBDMnH2TvrddhqTji5iZAMZ3RXs2NLpSnhH"
+func main() {
```
danlaine, line 50: nit: can `main` call a function `run()` instead of having application logic inside `main`? This makes it possible to just have one place where we do `os.Exit(1)` (namely, if `run` returns an error) instead of several places. See [here](https://github.com/ava-labs/avalanche-testing-new/blob/main/main.go#L56) for an example

vms/subnet.go
```diff
@@ -0,0 +1,364 @@
...
+					return ctx.Err()
+				case <-time.After(statusCheckTimeout):
+				}
+				status, err := nClient.GetBlockchainStatus(blockchainID.String())
+				if err != nil {
+					return fmt.Errorf("error querying blockchain status: %w", err)
+				}
+				if status == platformvm.Validating {
+					// after the first acceptance, next nodes probably don't need to check that long anymore
+					statusCheckTimeout = apiRetryFreq
```
danlaine, line 326: why do we use 2 different wait durations?

vms/subnet.go
```diff
@@ -0,0 +1,364 @@
...
+	allNodes       map[string]node.Node
+	rSubnetID      ids.ID
+}
+
+// SetupSubnet creates the necessary transactions to create a subnet for a given custom VM.
+// It then also waits until all these transactions are confirmed on all nodes.
+// Finally it creates a blockchain and makes sure all nodes of the given network
+// are validating this blockchain.
+// It requires a `privateKey` in order to issue the necessary transactions
+func SetupSubnet(
```
danlaine, line 45:
```suggestion
func CreateSubnetAndBlockchain(
```

danlaine (changes requested):

local/network.go
```diff
@@ -293,6 +300,60 @@ func NewDefaultConfig(binaryPath string) network.Config {
...
+		// If they are the same, we assume the user keeps it there, so we don't do nothing.
+		// Otherwise we copy the binary into the avalanchego plugins directory.
+		// TODO: This is a bit unstable:
+		// * no version control (needed?)
+		// * binaries just get overwritten
+		if pluginDir != vmBinDir {
+			if err := utils.CopyFile(v.Path, pluginDir); err != nil {
+				return network.Config{}, fmt.Errorf("failed to copy binary %s to pluginsDir %s: %w", v.Path, pluginDir, err)
+			}
+		}
```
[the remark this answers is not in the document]
holisticode: I agree much better

programs/local/customvm/main.go
```diff
@@ -0,0 +1,199 @@
...
+	watchShutdown := utils.WatchShutdownSignals(log, nw.Stop)
+
+	healthyChan := nw.Healthy(ctx)
+	log.Info("waiting for all nodes to report healthy...")
+	if err := <-healthyChan; err != nil {
+		return err
+	}
+	// use a new timed context as we need to wait for the validators validation start time
+	subnetCtx, subnetCancel := context.WithTimeout(ctx, subnetTimeout)
+	defer subnetCancel()
```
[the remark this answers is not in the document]
holisticode: We already have these named objects earlier, and I am explicitly not wanting to re-use the same context, but a new one.

vms/subnet.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+	return g.Wait()
+}
+
+// isSubnetInList returns an error if the given subnet is not in the client's list
+func isSubnetInList(txPChainClient platformvm.Client, rSubnetID ids.ID) error {
+	// confirm created subnet appears in subnet list
+	if _, err := txPChainClient.GetSubnets([]ids.ID{rSubnetID}); err != nil {
+		return fmt.Errorf("subnet not found: %w", err)
+	}
```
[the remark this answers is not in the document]
holisticode: I assumed we don't _need_ to, but it's probably better

vms/subnet.go
holisticode: Because until the nodes are validating there is a longer wait. Until the first node has been recognized to validate though, the other nodes should get validating probably be faster (i.e. at the same time actually), so we don't need long intervals anymore.

vms/subnet.go
```diff
@@ -0,0 +1,364 @@
...
+					status, err := client.GetTxStatus(txID, true)
+					if err != nil {
+						return err
+					}
+					if status.Status == platformvm.Committed {
+						args.log.Debug("add subnet validator (%s) tx (%s) accepted", nID, txID)
+						return nil
+					}
+					args.log.Debug("waiting for add subnet validator (%s) tx (%s) to be accepted", nID, txID)
+				}
```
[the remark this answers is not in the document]
holisticode: Created a `AwaitedAllNodesPChainTxAccepted` in the utils package.

vms/subnet.go
```diff
@@ -0,0 +1,345 @@
...
+	// but because it's an array let's be thorough
+	found := false
+	for _, api := range subnetAPIs {
+		if api.ID == rSubnetID {
+			found = true
+		}
+	}
+	if !found {
+		return errors.New("subnet not in returned list")
+	}
```
danlaine, line 186: Sorry, I was wrong before -- we don't need to check that the subnet ID is in the returned list since we're filtering by the subnet ID already. We just have to assert that the length of the returned list is 1

danlaine (changes requested):

programs/local/customvm/main.go
```diff
@@ -0,0 +1,208 @@
...
+	SubnetIDs    []string `json:"subnetIds"`
+	VMIDs        []string `json:"vmIds"`
+}
+
+// Shows example usage of the Avalanche Network Runner with custom VMs.
+// Creates a local five node Avalanche network all starting an avalanchego binary
+// and a provided custom VM binary. Waits for all nodes to become healthy.
+// The network runs until the user provides a SIGINT or SIGTERM.
+// Example of how to run this:
+// go run programs/local/customvm/main.go --vm-path "/path/to/vm/binary" --genesis-path "/path/to/genesis/file" --subnet-ids "24tZhrm8j8GCJRE9PomW8FaeqbgGS4UAQjJnqqn8pq5NwYSYV1" --vm-ids "tGas3T58KzdjLHhBDMnH2TvrddhqTji5iZAMZ3RXs2NLpSnhH"
```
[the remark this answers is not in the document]
holisticode: not much, but updated

programs/local/customvm/main.go
holisticode: Because then the flag definition kinda gets confusing.

Now, we have e.g. `--vm-ids="1,2" --vm-paths="/path/1","/path/2"` etc.
The way you suggest, we would have to do something like `--vms="1 /path/1, 2 /path/2" ` and then it's not clear to the user for each vm which of the cli param is which.

For the time being the default use case will be one vm only anyway, which results in the neat `--vm-ids="1", --vm-paths="/path/1"`

PR #70, opened 2022-01-03
fix local node config to allow unmarshaling from json

danlaine:


gyuho (approved):
nice! This resolves my issue. lgtm

PR #72, opened 2022-01-06
CHS 271 Flags

holisticode:
Closes https://ava-labs.atlassian.net/browse/CHS-271 

This PR adds flags to `network.Config` and `node.Config`.
Flags in `node.Config` override flags in `network.Config` if they are the same.

### DIFF
--- a/k8s/network_test.go
+++ b/k8s/network_test.go
@@ -380,6 +380,38 @@ func TestWrongNetworkConfigs(t *testing.T) {
 	}
 }
 
+// TestFlags tests that we can pass flags through the network.Config
+// but also via node.Config and that the latter overrides the former
+// if same keys exist.
+func TestFlags(t *testing.T) {
+	assert := assert.New(t)
+	networkConfig := defaultTestNetworkConfig(t)
+
+	networkConfig.Flags = map[string]interface{}{
+		"test-network-config-flag": "something",
+		"common-config-flag":       "should not be added",
+	}
+	for i := range networkConfig.NodeConfigs {
+		v := &networkConfig.NodeConfigs[i]
+		v.Flags = map[string]interface{}{
+			"test-node-config-flag":  "node",
+			"test2-node-config-flag": "config",
+			"common-config-flag":     "this should be added",
+		}
+	}
+	_, err := newTestNetworkWithConfig(networkConfig)
+	assert.NoError(err)
+	// after creating the network, one flag should have been overridden by the node configs
+	for _, n := range networkConfig.NodeConfigs {
+		assert.True(len(n.Flags) == 4)
+		assert.Contains(n.Flags, "test-network-config-flag")
+		assert.Contains(n.Flags, "common-config-flag")
+		assert.Equal(n.Flags["common-config-flag"], "this should be added")
+		assert.Contains(n.Flags, "test-node-config-flag")
+		assert.Contains(n.Flags, "test2-node-config-flag")
+	}
+}
+
 // TestImplSpecificConfigInterface checks incorrect type to interface{} ImplSpecificConfig
 // This is adapted from the local test suite
 func TestImplSpecificConfigInterface(t *testing.T) {
--- a/k8s/network_test.go
+++ b/k8s/network_test.go
@@ -403,12 +403,15 @@ func TestFlags(t *testing.T) {
 	assert.NoError(err)
 	// after creating the network, one flag should have been overridden by the node configs
 	for _, n := range networkConfig.NodeConfigs {
-		assert.True(len(n.Flags) == 4)
+		assert.Len(n.Flags, 4)
 		assert.Contains(n.Flags, "test-network-config-flag")
+		assert.Equal(n.Flags["test-network-config-flag"], "something")
 		assert.Contains(n.Flags, "common-config-flag")
 		assert.Equal(n.Flags["common-config-flag"], "this should be added")
 		assert.Contains(n.Flags, "test-node-config-flag")
+		assert.Equal(n.Flags["test-node-config-flag"], "node")
 		assert.Contains(n.Flags, "test2-node-config-flag")
+		assert.Equal(n.Flags["test2-node-config-flag"], "config")
 	}
 	err = nw.Stop(context.Background())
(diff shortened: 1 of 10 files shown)


local/network.go
```diff
@@ -379,6 +392,9 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
 		fmt.Sprintf("--%s=%s", config.BootstrapIDsKey, ln.bootstrapIDs),
 	}
 
+	for flagName, flagVal := range nodeConfig.Flags {
+		flags = append(flags, fmt.Sprintf("--%s=%v", flagName, flagVal))
+	}
```
danlaine, line 397: We should log a warning if the user overrides `config.NetworkNameKey`, `config.BootstrapIPsKey` or `config.BootstrapIDsKey` because not using the ones provided by the network runner is probably going to result in the node not working. We could just disallow the usage of those flags, too. What do you think @holisticode ?

network/node/node.go
```diff
@@ -47,6 +47,9 @@ type Config struct {
 	ConfigFile string `json:"configFile"`
 	// May be nil.
 	CChainConfigFile string `json:"cChainConfigFile"`
+	// Flags can hold additional flags for the node.
+	// It can be empty.
+	Flags map[string]interface{} `json:"flags"`
```
danlaine, line 52: Can you document here that the precedence from highest to lowest is:
1, Config file
2. Flags defined in network.Config
3. Flags defined in node.Config

network/config.go
```diff
@@ -79,6 +79,9 @@ type Config struct {
 	Name string `json:"name"`
 	// Backend specifies the backend for the network
 	Backend Backend `json:"backend"`
+	// Flags can hold additional flags for avalanchego nodes.
+	// It can be empty.
+	Flags map[string]interface{}`json:"flags"`
```
danlaine, line 84: Can you document here that the precedence from highest to lowest is:
1, Config file
2. Flags defined in network.Config
3. Flags defined in node.Config

k8s/utils.go
```diff
@@ -146,9 +146,24 @@ func validateObjectSpec(k8sobj ObjectSpec) error {
...
+	nodeConfigs := params.conf.NodeConfigs
 	var beacons, nonBeacons []*k8sapi.Avalanchego
 	names := make(map[string]struct{})
+
+	for flagName, flagVal := range params.conf.Flags {
+		for i := range params.conf.NodeConfigs {
+			n := &params.conf.NodeConfigs[i]
+			// Do not overwrite flags described in the nodeConfig
+			if len(n.Flags) == 0 {
+				n.Flags = make(map[string]interface{})
```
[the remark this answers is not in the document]
holisticode: Nice catch, thought it was being handled via `buildNodeEnv` but that handles only stuff from the config file.

local/network.go
holisticode: We could have an array of unsupported flags, and if any of these are provided we just exit.

local/network_test.go
```diff
@@ -612,6 +612,86 @@ func TestGetAllNodes(t *testing.T) {
...
+			"test2-node-config-flag": "config",
+			"common-config-flag":     "this should be added",
+		}
+	}
+	nw, err := newNetwork(logging.NoLog{}, networkConfig, newMockAPISuccessful, newMockProcessSuccessful)
+	assert.NoError(err)
+	// after creating the network, one flag should have been overridden by the node configs
+	for _, n := range networkConfig.NodeConfigs {
+		assert.True(len(n.Flags) == 4)
+		assert.Contains(n.Flags, "test-network-config-flag")
```
felipemadero, line 640: probably use assert.Equal for all flags and check all expected values in all cases? no sure it is checking that network config flags are beeing copied

local/network_test.go
```diff
@@ -612,6 +612,86 @@ func TestGetAllNodes(t *testing.T) {
...
+			"test2-node-config-flag": "config",
+			"common-config-flag":     "this should be added",
+		}
+	}
+	nw, err = newNetwork(logging.NoLog{}, networkConfig, newMockAPISuccessful, newMockProcessSuccessful)
+	assert.NoError(err)
+	// after creating the network, only node configs should exist
+	for _, n := range networkConfig.NodeConfigs {
+		assert.True(len(n.Flags) == 3)
+		assert.NotContains(n.Flags, "test-network-config-flag")
```
felipemadero, line 664: same as before. check for all values probably. and NotContains check seems not to be needed

local/network_test.go
```diff
@@ -612,6 +612,86 @@ func TestGetAllNodes(t *testing.T) {
...
+	}
+	for i := range networkConfig.NodeConfigs {
+		v := &networkConfig.NodeConfigs[i]
+		v.Flags = nil
+	}
+	nw, err = newNetwork(logging.NoLog{}, networkConfig, newMockAPISuccessful, newMockProcessSuccessful)
+	assert.NoError(err)
+	// after creating the network, only flags from the network config should exist
+	for _, n := range networkConfig.NodeConfigs {
+		assert.True(len(n.Flags) == 2)
```
felipemadero: same as before

k8s/network_test.go
```diff
@@ -380,6 +380,85 @@ func TestWrongNetworkConfigs(t *testing.T) {
...
+			"test-node-config-flag":  "node",
+			"test2-node-config-flag": "config",
+			"common-config-flag":     "this should be added",
+		}
+	}
+	nw, err := newTestNetworkWithConfig(networkConfig)
+	assert.NoError(err)
+	// after creating the network, one flag should have been overridden by the node configs
+	for _, n := range networkConfig.NodeConfigs {
+		assert.True(len(n.Flags) == 4)
```
felipemadero, line 406: same comments as per local example

felipemadero (commented):
lgtm

local/network.go
danlaine: @holisticode Can you please add a warning log that's printed when the user sets one of the above flags? Other than that, LGTM :)

holisticode: Done

danlaine (approved):

PR #74, opened 2022-01-07
Generate random port

holisticode:
As the actual node starts later than when the port is assigned, it can happen that the os assigns the same port twice.

Try with a random port and verification instead.


local/network.go
```diff
@@ -35,6 +36,8 @@ const (
 	stopTimeout           = 30 * time.Second
 	healthCheckFreq       = 3 * time.Second
 	defaultNumNodes       = 5
+	maxPort               = 65535
```
danlaine, line 39:
```suggestion
	maxPort               = math.MaxUint16
```

local/network.go
```diff
@@ -342,12 +345,15 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
...
-		l, err := net.Listen("tcp", ":0")
-		if err != nil {
-			return nil, fmt.Errorf("could not get free port: %w", err)
+		for {
+			r := rand.Intn(maxPort-minPort) + minPort
+			l, err := net.Listen("tcp", fmt.Sprintf(":%d", r))
+			if err == nil {
+				apiPort = uint16(l.Addr().(*net.TCPAddr).Port)
+				_ = l.Close()
+				break
```
danlaine, line 354: 1. Let's add some comments explaining what we're doing
2. As you suggested on Slack, let's use a context with a timeout and a select statement to make sure this doesn't loop forever

local/network.go
```diff
@@ -342,12 +345,15 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
 		}
 	} else {
 		// Get a free port for the API port
-		l, err := net.Listen("tcp", ":0")
-		if err != nil {
-			return nil, fmt.Errorf("could not get free port: %w", err)
+		for {
+			r := rand.Intn(maxPort-minPort) + minPort
```
danlaine, line 349:
```suggestion
			port := rand.Intn(maxPort-minPort+1) + minPort
```

danlaine (changes requested):

danlaine (approved):

PR #75, opened 2022-01-10
refactor flag passing

danlaine:


k8s/utils.go
```diff
@@ -29,74 +30,64 @@ func convertKey(key string) string {
...
-		} else {
-			// new var
-			v := corev1.EnvVar{
-				Name:  convertKey(key),
-				Value: val.(string),
-			}
-			env = append(env, v)
-		}
+
+	// For each config key, convert it to the format
```
holisticode, line 68: This comment is at the wrong line

network/config.go
```diff
@@ -36,17 +36,16 @@ type AddrAndBalance struct {
 	Balance uint64
 }
 
-// Backend enumerator type
+// The type of network runner
```
holisticode, line 39: Shouldn't comments of exported items start with the exported name?

network/config.go
```diff
@@ -85,12 +83,18 @@ type Config struct {
...
+	// Config flags may also be passed in a node's config struct
+	// or config file.
+	// The precedence of flags handling is, from highest to lowest:
+	// 1. Flags defined in a node's node.Config
+	// 2. Flags defined in a network's network.Config
+	// 3. Flags defined in a node's config file
+	// For example, if a network.Config has flag W set to X,
+	// and a node within that network has flag W set to Y,
+	// and the node's config file has flag W set to Z,
+	// then the node will be started with flag W set to X.
```
holisticode, line 97: It would be set to `Y` - unless I don't understand line 95?

holisticode (changes requested):
some nits

network/config.go
danlaine: I'm not super committed to that approach -- sometimes I think it makes the comment sound clunky. In any case, changed this comment to start with `Backend`

k8s/utils.go
danlaine: Moved

network/config.go
danlaine: woops! fixed, good catch.

holisticode (approved):

PR #77, opened 2022-01-11
Revert "update avalanchego dependency"

danlaine:
Reverts ava-labs/avalanche-network-runner#76


danlaine:
accidentally updated avalanchego dependency to a non-public version

PR #78, opened 2022-01-11
Abigen integration

felipemadero:


api/eth_client.go
```diff
@@ -160,3 +169,84 @@ func (c *ethClient) SuggestGasPrice(ctx context.Context) (*big.Int, error) {
 	}
 	return c.client.SuggestGasPrice(ctx)
 }
+
+func (c *ethClient) AcceptedCodeAt(ctx context.Context, account common.Address) ([]byte, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if err := c.connect(); err != nil {
```
holisticode, line 176: Why do we have to always connect first? Where are we connecting to?

holisticode (approved):
Lgtm, check the question

api/eth_client.go
danlaine: @felipemadero 

Can you please factor out this bit:

```
c.lock.Lock()
	defer c.lock.Unlock()
	if err := c.connect(); err != nil {
		return nil, err
	}
``` ?

felipemadero: > Why do we have to always connect first? Where are we connecting to?

As there were some unexpected disconnection issues, it was decided to just connect in every call. It is making a new websocket connection to coreth eth API.

danlaine: @felipemadero We never set `client` to nil, though, so `connect` only ever does something the first time it runs. Even if there in an error with the connection. Am I missing something?

felipemadero: @danlaine is just what you said. It's just for the first connection. Probably we can leave this logic here because we can't create a connection on node initialization. Or we can add a Connect() method that is going to be called by network runner after Healthy check (or after node initialization, introducing some logic for this into network runner -operations that network runner do on nodes after they -each one- initialize- ).
Regarding disconnections, will check is websocket disconnection is happening in tests (usually dynamic feeds is good case), and see if with the library we can solve the issue.

felipemadero: Anyway that two things are for another ticket in my opinion. Also, there may be the need to have a ticket to make the lib 1-to-1 with coreth ethclient, instead of being adding methods each time it is needed. Probably this ticket for that.

danlaine (approved):

PR #83, opened 2022-01-18
Colorize child process output

holisticode:
Closes https://ava-labs.atlassian.net/browse/CHS-278

### DIFF
--- a/local/network.go
+++ b/local/network.go
@@ -1,11 +1,11 @@
 package local
 
 import (
-	"bufio"
 	"context"
 	"embed"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"os/exec"
@@ -128,6 +128,7 @@ func init() {
 	}
 }
 
+// NodeProcessCreator is an interface for new node process creation
 type NodeProcessCreator interface {
 	NewNodeProcess(config node.Config, args ...string) (NodeProcess, error)
 }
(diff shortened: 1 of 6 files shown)


utils/color.go
```diff
@@ -0,0 +1,71 @@
...
+	logging.Yellow,
+}
+
+type color struct {
+	lock       sync.Mutex
+	usedColors []logging.Color
+}
+
+// ColorPicker allows to assign a color to different clients
+var ColorPicker = &color{
```
danlaine, line 29: Let's not use a global variable here. We could have a function `GetNextColor(i int)` which returns the color at `supportedColors[i%len(supportedColors)`. (That might be the easiest.) Or we could have a function `NewColorPicker` which returns a color picker struct.

local/network.go
```diff
@@ -127,25 +128,58 @@ func init() {
...
+type nodeProcessCreator struct {
+	color logging.Color
+}
+
+// default variables pointing to the default functions to get the os output files
+// we can replace these in testing (see TestChildCmdRedirection in local/network_test.go)
+var (
+	stdOutFunc = getDefaultStdOut
+	stdErrFunc = getDefaultStdErr
+)
```
danlaine, line 145: I think we can avoid having the globals/overwriting them in tests by adding additional fields to `nodeProcessCreator`.

Namely, we can add fields `stdout io.Writer` and `stderr io.Writer` to control where stdout and stderr are written for testing

local/network.go
```diff
@@ -127,25 +128,58 @@ func init() {
...
 }
 
+// NodeProcessCreator is an interface for new node process creation
 type NodeProcessCreator interface {
 	NewNodeProcess(config node.Config, args ...string) (NodeProcess, error)
 }
 
-type nodeProcessCreator struct{}
+type nodeProcessCreator struct {
+	color logging.Color
```
danlaine, line 137: comment please

utils/color.go
```diff
@@ -0,0 +1,71 @@
...
+	go func(scanner *bufio.Scanner) {
+		// we should not need any go routine control here:
+		// when the program exits, `Scan()` will hit an EOF and return false,
+		// and therefore the routine terminates
+		for scanner.Scan() {
+			txt := assignedColor.Wrap(fmt.Sprintf("[%s] %s\n", wrapText, scanner.Text()))
+			if _, err := writer.Write([]byte(txt)); err != nil {
+				// TODO better handling required? Failing and returning doesn't seem quite right either
+				fmt.Println("failed to write wrapped text to writer")
+			}
```
danlaine, line 66: I think we can just ignore errors on write since I don't think they should ever really happen

danlaine (changes requested):

utils/color.go
holisticode: I think the use of a global variable is justified here. Specifically, the suggested change significantly deteriorates usage for clients. Also this is a fairly uncritical part of the code.

local/network.go
holisticode: Much better and elegant, also uses way less code

utils/color.go
holisticode: I agree

PR #88, opened 2022-01-27
Put all local network runner logs/files under one directory

holisticode:
Closes https://ava-labs.atlassian.net/browse/FRAM-216

Also "standardized" `localNetwork` receiver instances to be all `ln` against with mixed `net`

### DIFF
--- a/local/network.go
+++ b/local/network.go
@@ -69,7 +69,8 @@ type localNetwork struct {
 	nodes map[string]*localNode
 	// List of nodes that new nodes will bootstrap from.
 	bootstrapIPs, bootstrapIDs beaconList
-	// rootDir is the root directory of all node process directories
+	// rootDir is the root directory under which we write all node
+	// logs, databases, etc.
 	rootDir string
 }
 
@@ -241,7 +242,7 @@ func newNetwork(
 		}
 	}
 
-	net.rootDir, err = os.MkdirTemp("", "avalanchego-network-runner-*")
+	net.rootDir, err = os.MkdirTemp("", "avalanche-network-runner-*")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


local/network.go
```diff
@@ -323,12 +332,15 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
 		return nil, fmt.Errorf("repeated node name %s", nodeConfig.Name)
 	}
 
-	// [tmpDir] is where this node's config file, C-Chain config file,
+	if ln.rootDir == "" {
+		ln.log.Warn("no network root directory defined; will create this node's runtime directory in working directory")
+	}
```
danlaine, line 337: This shouldn't ever happen, right?

local/network.go
```diff
@@ -362,6 +374,7 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
 
 	// Use random free API port, unless given in config
 	var apiPort uint16
+	var err error
```
danlaine, line 377:
```suggestion
	var (
		// Use random free API port, unless given in config
		apiPort uint16
		err     error
	)
```

local/network.go
```diff
@@ -69,6 +69,8 @@ type localNetwork struct {
 	nodes map[string]*localNode
 	// List of nodes that new nodes will bootstrap from.
 	bootstrapIPs, bootstrapIDs beaconList
+	// rootDir is the root directory of all node process directories
```
danlaine, line 72:
```suggestion
	// rootDir is the root directory under which we write all node
	// logs, databases, etc.
```

local/network.go
```diff
@@ -238,6 +241,11 @@ func newNetwork(
 		}
 	}
 
+	net.rootDir, err = os.MkdirTemp("", "avalanchego-network-runner-*")
```
danlaine, line 244:
```suggestion
	net.rootDir, err = os.MkdirTemp("", "avalanche-network-runner-*")
```
to be consistent with namind

danlaine (changes requested):
A few small nits by otherwise LGTM :+1:

local/network.go
holisticode: Not if used in the normal flow, but just to make sure, as this is in `addNode`, and someone could be calling it via `AddNode` without the default config?

danlaine (approved):

danlaine (approved):

PR #90, opened 2022-02-04
Fixing API and P2P port flag from nodeConfig

rajranjan0608:
## Background

I was working on setting up **Avalanche Network Runner**, and got through `indepth` example. Custom API port works fine for 5 default nodes (whose config files are present) but when we add another node in the network, `network.go` ignore API Port flag (in `addNode()` function). It only checks for API port from config file.

Here in this [line](https://github.com/ava-labs/avalanche-network-runner/blob/813da3d67099f5f9683704531257d2435e982adb/local/network.go#L412) we are checking `config.HTTPPortKey` only from `configFile`. But we are ignoring it, if it’s coming as a flag from `nodeConfig`. This can be solved if we can also check flags, like

```go
    if apiPortIntf, ok := nodeConfig.Flags[config.HTTPPortKey]; ok {
```

The flag value is being read indeed [here](https://github.com/ava-labs/avalanche-network-runner/blob/813da3d67099f5f9683704531257d2435e982adb/local/network.go#L455), and so it should print as well. But it’s not being used while assigning API port.

## Solution

Once the flags are being fetched with keys and values, we will check again, if there is any API or P2P port flag coming from `nodeConfig`. If it's there, then we can append the updated port flags.

PS - Please let me know about the specific naming conventions you are using. I have tried to keep it similar to other parts of the code.

### DIFF
--- a/local/network.go
+++ b/local/network.go
@@ -404,16 +404,22 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
 		}
 	}
 
-	// Use random free API port, unless given in config
+	// Use random free API port, unless given in config file or node config flag
 	var (
 		apiPort uint16
 		err     error
 	)
 	if apiPortIntf, ok := configFile[config.HTTPPortKey]; ok {
-		if apiPortFromConfig, ok := apiPortIntf.(float64); ok {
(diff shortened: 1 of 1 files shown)


felipemadero (approved):
LGTM

PR #94, opened 2022-02-13
Add support for specifying network root directory

dizzyd:
By default the avalanche network runner creates temporary directories that are not accessible/visible to the caller. This makes it hard to pre-populate configuration (such as chain config) for custom VMs. This PR introduces the ability to pass in a root directory for the network and also ensures nodes can still start up even if the directory already exists.


danlaine:
Opened [this PR](https://github.com/ava-labs/avalanche-network-runner/pull/95) which has the commits in here plus a few small fixes

PR #95, opened 2022-02-14
Allow user to specify network root directory 

danlaine:
This PR allows the user to specify where a local network's files are written by default.

This PR is branched off of [this one](https://github.com/ava-labs/avalanche-network-runner/pull/94).
@dizzyd I added one more commit with some nits and test fixes. Thank you for the PR!


local/network.go
```diff
@@ -254,10 +267,16 @@ func newNetwork(
...
-	if err != nil {
-		return nil, err
+
+	if dir == "" {
+		net.rootDir, err = os.MkdirTemp("", "avalanche-network-runner-*")
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		net.rootDir = dir
```
holisticode, line 277: Should we check if this dir actually exists, and maybe even, if not, create it (or warn and exit if it doesn't exist)?

holisticode (commented):
Looks good to me, a minor question

holisticode (approved):

Issue #96, opened 2022-02-16
Error running examples/local/fivenodenetwork


manishmarathe:
The --chain-config-dir param needs a directory not the actual file path. So in local/network.go, in writeFiles method, after writing the chain config file, use a different variable ppath to pass the directory for this particular parameter. Flexibility added for other params in the writeFiles function. This was the reason the network failed to get created.

It seems I don't have permissions to push to remote and generate a PR.

Here is the patch:

```go
index f766673..5f1c4ba 100644
--- a/local/network.go
+++ b/local/network.go
@@ -787,46 +787,52 @@ func (ln *localNetwork) buildFlags(
 // It returns flags used to point to those files.
 func writeFiles(genesis []byte, nodeRootDir string, nodeConfig *node.Config) ([]string, error) {
        type file struct {
-               path    string
+               ppath   string
+               fpath   string
                pathKey string
                val     []byte
        }
        files := []file{
                {
-                       path:    filepath.Join(nodeRootDir, stakingKeyFileName),
+                       ppath:    filepath.Join(nodeRootDir, stakingKeyFileName),
+                       fpath:    filepath.Join(nodeRootDir, stakingKeyFileName),
                        pathKey: config.StakingKeyPathKey,
                        val:     []byte(nodeConfig.StakingKey),
                },
                {
-                       path:    filepath.Join(nodeRootDir, stakingCertFileName),
+                       ppath:   filepath.Join(nodeRootDir, stakingCertFileName),
+                       fpath:   filepath.Join(nodeRootDir, stakingCertFileName),
                        pathKey: config.StakingCertPathKey,
                        val:     []byte(nodeConfig.StakingCert),
                },
                {
-                       path:    filepath.Join(nodeRootDir, genesisFileName),
+                       ppath:   filepath.Join(nodeRootDir, genesisFileName),
+                       fpath:   filepath.Join(nodeRootDir, genesisFileName),
                        pathKey: config.GenesisConfigFileKey,
                        val:     genesis,
                },
        }
        if len(nodeConfig.ConfigFile) != 0 {
                files = append(files, file{
-                       path:    filepath.Join(nodeRootDir, configFileName),
+                       ppath:   filepath.Join(nodeRootDir, configFileName),
+                       fpath:   filepath.Join(nodeRootDir, configFileName),
                        pathKey: config.ConfigFileKey,
                        val:     []byte(nodeConfig.ConfigFile),
                })
        }
[...]

PR #97, opened 2022-02-16
fix broken chain config

danlaine:
Thanks @manishmarathe for [flagging this](https://github.com/ava-labs/avalanche-network-runner/issues/96)

### DIFF
--- a/local/network.go
+++ b/local/network.go
@@ -46,6 +46,8 @@ var (
 		config.BootstrapIPsKey: {},
 		config.BootstrapIDsKey: {},
 	}
+	chainConfigSubDir  = "chainConfigs"
+	cChainConfigSubDir = filepath.Join(chainConfigSubDir, "C")
 )
 
 // network keeps information uses for network management, and accessing all the nodes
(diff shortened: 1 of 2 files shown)


local/network.go
```diff
@@ -787,46 +787,52 @@ func (ln *localNetwork) buildFlags(
 // It returns flags used to point to those files.
 func writeFiles(genesis []byte, nodeRootDir string, nodeConfig *node.Config) ([]string, error) {
 	type file struct {
-		path    string
+		ppath   string
```
holisticode, line 790: Could you please add some comments what these two names stand for?

local/network.go
```diff
@@ -787,46 +787,52 @@ func (ln *localNetwork) buildFlags(
...
 			pathKey: config.ChainConfigDirKey,
 			val:     []byte(nodeConfig.CChainConfigFile),
 		})
 	}
 	flags := []string{}
 	for _, f := range files {
-		flags = append(flags, fmt.Sprintf("--%s=%s", f.pathKey, f.path))
-		if err := createFileAndWrite(f.path, f.val); err != nil {
-			return nil, fmt.Errorf("couldn't write file at %q: %w", f.path, err)
+		flags = append(flags, fmt.Sprintf("--%s=%s", f.pathKey, f.ppath))
```
holisticode, line 833: Ah so this means that the avalanchego node needs the flag to the dir, not the file? Kinda weird though too

holisticode (approved):

PR #99, opened 2022-02-21
C-Chain config: enable Apricot phases 3, 4, 5

lebdron:


patrick-ogrady:
Nice find! https://github.com/ava-labs/coreth/blob/master/plugin/evm/vm_test.go#L84

gyuho (approved):

PR #102, opened 2022-02-24
*: support RPC server

gyuho:
**What does `network-runner` do?** The primary focus of [`network-runner`](https://github.com/ava-labs/avalanche-network-runner) is to create a local network, as a test framework for local development.

**Why `network-runner` as a binary?** Previously, each team was still required to write a substantial amount of Go code to integrate with `network-runner`. And the circular dependency on `avalanchego` made it unusable within `avalanchego` itself. Using `network-runner` as a binary (rather than Go package) eliminates the complexity of such dependency management.

**Why `network-runner` needs RPC server?** `network-runner` needs to provide more complex workflow such as replace, restart, inject fail points, etc.. The RPC server will expose basic node operations to enable a separation of concerns such that one team develops test framework, and the other writes test cases and its controlling logic.

**Why gRPC?** The RPC server leads to more modular test components, and gRPC enables greater flexibility. The protocol buffer increases flexibility as we develop more complicated test cases. And gRPC opens up a variety of different approaches for how to write test controller (e.g., Rust). See [`rpcpb/rpc.proto`](./rpcpb/rpc.proto) for service definition.

**Why gRPC gateway?** [gRPC gateway](https://grpc-ecosystem.github.io/grpc-gateway/) exposes gRPC API via HTTP, without us writing any code. Which can be useful if a test controller writer does not want to deal with gRPC.

Signed-off-by: Gyuho Lee <gyuho.lee@avalabs.org>


gyuho:
Basically, porting features from https://github.com/gyuho/avax-tester.

Requirement for avalanche platform team to run e2e tests without circular dependencies in `avalanchego`.

.github/workflows/build-test-release.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Run e2e tests
+        shell: bash
+        run: scripts/tests.e2e.sh 1.7.4 1.7.5
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          distribution: goreleaser
+          version: latest
+          args: release --rm-dist
+        env:
```
patrick-ogrady, line 31: Do we have a secret populated?

patrick-ogrady (approved):
WCGW

.github/workflows/build-test-release.yml
gyuho: > GitHub automatically creates a unique GITHUB_TOKEN secret to use in your workflow. You can use the GITHUB_TOKEN to authenticate in a workflow run.

PR #112, opened 2022-03-18
Use arbitrary number of nodes in control command

felipemadero:


gyuho (approved):

client/client.go
```diff
@@ -30,7 +30,7 @@ type Config struct {
 
 type Client interface {
 	Ping(ctx context.Context) (*rpcpb.PingResponse, error)
-	Start(ctx context.Context, execPath string, opts ...OpOption) (*rpcpb.StartResponse, error)
+	Start(ctx context.Context, execPath string, numNodes uint32, opts ...OpOption) (*rpcpb.StartResponse, error)
```
gyuho, line 33: On second thought, can we make `numNodes` as `OpOption`?

`WithNumNodes(n uint32)` can set a non-default number of node?

gyuho: That way, we. don't need to break the client interface

local/network.go
```diff
@@ -325,6 +326,37 @@ func NewDefaultConfig(binaryPath string) network.Config {
...
+		for i := 0; i < toAdd; i++ {
+			nodeConfig := refNodeConfig
+			stakingCert, stakingKey, err := staking.NewCertAndKeyBytes()
+			if err != nil {
+				return netConfig, fmt.Errorf("couldn't generate staking Cert/Key: %w", err)
+			}
+			nodeConfig.StakingKey = string(stakingKey)
+			nodeConfig.StakingCert = string(stakingCert)
+			// replace api port in refNodeConfig.ConfigFile
+			apiPort, err := getFreePort()
```
holisticode, line 344: don't we need to also look for a free P2P port?

felipemadero: No. Reason is that the reference node config sets only API port. I need to replace that conf with a new port in flags. Regarding the P2P port, as is not defined in any conf, it will be automatically generated.

gyuho (approved):

server/network.go
```diff
@@ -72,6 +75,9 @@ func newNetwork(execPath string, rootDataDir string, whitelistedSubnets string,
 		nodeNames[i] = nodeName
 		cfg.NodeConfigs[i].Name = nodeName
 
+		// drop any flags received in original conf
+		cfg.NodeConfigs[i].Flags = nil
```
danlaine, line 79: Why do we do this?

danlaine (approved):

server/network.go
felipemadero: Probably not needed. I wanted to keep the previous behaviour of the network created by the server, which was to let the runner create the ports nums when executing addNode (as the node config in server does not have API port, no flag was needed).

felipemadero: Will remove this. Not needed at this moment.

PR #122, opened 2022-04-01
Improve unit tests time

felipemadero:


danlaine:
@felipemadero 
I looked into it a bit and it seems that calls to `testNetworkConfig()` are what are taking a lot of time because that function creates TLS keys/certs. Instead of creating new key/cert pairs on the fly, we should just read in constant ones from files. This will make the tests faster and less CPU intensive.  Can you please add that? I think if we do that we can probably get the amount of time the unit tests take down below 30s.

danlaine (approved):

PR #126, opened 2022-04-17
fix AttachPeer so ID() returns the ID of the test peer and not the ID…

danlaine:
… of the node it's attached to


aaronbuchwald (approved):
LGTM

gyuho (approved):

PR #127, opened 2022-04-17
Use existing ctx instead of background

aaronbuchwald:
This PR uses the already passed in context parameter instead of creating a new instance of background context within a helper function.

This may cause some changes in behavior where we used to rely on a two minute timeout, but I think this is better to implement using `context.WithTimeout` at the application level since this is very easy to do and allows the client to set it.

### DIFF
--- a/server/network.go
+++ b/server/network.go
@@ -12,7 +12,6 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/ava-labs/avalanche-network-runner/api"
 	"github.com/ava-labs/avalanche-network-runner/local"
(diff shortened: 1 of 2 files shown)


danlaine:
Yeah, I think a fixed timeout for becoming healthy doesn't make sense, since if bootstrapping takes > 2 minutes, it errors, which is obviously bad since bootstrapping may take that long on a network with non-trivial state.

danlaine (approved):

danlaine:
If I'm not mistaken, these requests are still subject to the request timeout (default 1 minute) defined [here](https://github.com/ava-labs/avalanche-network-runner/blob/main/cmd/avalanche-network-runner/control/control.go#L42), so that's something to be aware of. So if you use the `health` command without specifying a different timeout, it will error after 1 minute. Just something to be aware of.

gyuho:
Yeah in general this is a better approach. If 2-minute is not long enough for some cases, client can adjust.

gyuho (approved):

PR #128, opened 2022-04-20
First iteration for AddNode

holisticode:
Closes https://ava-labs.atlassian.net/browse/FRAM-296

WARNING: Creating this PR resulted in an *unexpected* merge conflict in a *generated* gRPC file `rpcpb/rpc.pb.go` during the necessary `git rebase` operation to fix the merge conflict. 

This was pretty surprising. 

The script `./scripts/genproto.sh` was then run again which overwrote the conflicting generated file. It might have fixed the underlying issue, but this PR should be carefully reviewed.

Also take the time to consider what exactly should be provided to a new independent node (whitelisted subnets, plugin dir, etc.).


client/client.go
```diff
@@ -194,6 +195,22 @@ func (c *client) Stop(ctx context.Context) (*rpcpb.StopResponse, error) {
 	return c.controlc.Stop(ctx, &rpcpb.StopRequest{})
 }
 
+func (c *client) AddNode(ctx context.Context, name string, execPath string, opts ...OpOption) (*rpcpb.AddNodeResponse, error) {
+	ret := &Op{}
+	ret.applyOpts(opts)
+
```
gyuho, line 201: Can we follow the pattern we have in `RestartNode` so optional fields are just set to null?

```go
	if ret.whitelistedSubnets != "" {
		req.WhitelistedSubnets = &ret.whitelistedSubnets
	}
	if ret.logLevel != "" {
		req.LogLevel = &ret.logLevel
	}
	if ret.rootDataDir != "" {
		req.RootDataDir = &ret.rootDataDir
	}
```

server/server.go
```diff
@@ -514,6 +515,64 @@ func (s *server) recvLoop(stream rpcpb.ControlService_StreamStatusServer) error
...
+	zap.L().Debug("received add node request", zap.String("name", req.Name))
+
+	if info := s.getClusterInfo(); info == nil {
+		return nil, ErrNotBootstrapped
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	nodeName := req.Name
```
gyuho, line 528: Can we error if the name is in conflict (e.g., `s.network.nodeInfos`)?

Can we error if the specified binary path does not exist?

client/client.go
holisticode: done

server/server.go
holisticode: I tested this and it fails anyway later when running `s.network.nw.AddNode`.
However it makes sense to check and fail early, so will change accordingly.

danlaine:
@holisticode  This doesn't seem to work just yet.

In one tab, do:
```
avalanche-network-runner server \
--log-level debug \
--port=":8080" \
--grpc-gateway-port=":8081"
```

In another, do:
```
avalanche-network-runner control start \
--log-level debug \               
--endpoint="0.0.0.0:8080" \
--avalanchego-path ${AVALANCHEGO_EXEC_PATH}
dialing endpoint "0.0.0.0:8080"
```

and then:
```
avalanche-network-runner control add-node --log-level=info --avalanchego-path $AVALANCHEGO_EXEC_PATH --node-name bob
```

This causes the client to crash.

Also, can we please add an E2E test for this and document this new functionality in the README?

server/server.go
```diff
@@ -514,6 +516,82 @@ func (s *server) recvLoop(stream rpcpb.ControlService_StreamStatusServer) error
...
+		return nil, fmt.Errorf("couldn't generate staking Cert/Key: %w", err)
+	}
+	nodeConfig := node.Config{
+		Name:               req.Name,
+		ConfigFile:         configFile,
+		ImplSpecificConfig: json.RawMessage(fmt.Sprintf(`{"binaryPath":"%s","redirectStdout":true,"redirectStderr":true}`, execPath)),
+		StakingKey:         string(stakingKey),
+		StakingCert:        string(stakingCert),
+	}
+	s.network.nodeNames = append(s.network.nodeNames, req.Name)
```
danlaine, line 575: I think we should do this after we've called `s.network.nw.AddNode` in case that errors

server/server.go
```diff
@@ -514,6 +516,82 @@ func (s *server) recvLoop(stream rpcpb.ControlService_StreamStatusServer) error
...
+	if info := s.getClusterInfo(); info == nil {
+		return nil, ErrNotBootstrapped
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	var logLevel, whitelistedSubnets, pluginDir string
+
+	nodeName := req.Name
```
danlaine, line 531: I don't think we need this variable

danlaine:
left 2 small comments, also bumping the above: `Also, can we please add an E2E test for this and document this new functionality in the README?`

server/server.go
holisticode: This is a debatable nit really, but ok.

danlaine: We should use one or the other but not both

danlaine (approved):

PR #129, opened 2022-04-20
add script to generate all network mocks

felipemadero:


scripts/genmocks.sh
```diff
@@ -0,0 +1,14 @@
...
+  echo "must be run from repository root"
+  exit 255
+fi
+
+mockery --dir api --name Client --output api/mocks/ --filename client.go
+mockery --dir api --name EthClient --output api/mocks/ --filename EthClient.go
+mockery --dir local --name NodeProcess --output local/mocks/ --filename node_process.go
+mockery --dir k8s --name dnsReachableChecker --output k8s/mocks/ --filename dns_checker.go --structname DnsReachableChecker
+
+echo "ALL SUCCESS"
```
holisticode, line 14: Nit: this will print "ALL SUCCESS" even if any of the above calls hits an error

holisticode (approved):

scripts/genmocks.sh
felipemadero: I understood the "set -e" prevents that

danlaine, line 14:
```suggestion
echo "Successfully generate mock files"
```

danlaine (approved):

PR #130, opened 2022-04-21
Add support for passing chain configs over RPC

darioush:
only works for C chain for now.


client/client.go
```diff
@@ -107,8 +107,9 @@ func (c *client) Start(ctx context.Context, execPath string, opts ...OpOption) (
 	ret.applyOpts(opts)
 
 	req := &rpcpb.StartRequest{
-		ExecPath: execPath,
-		NumNodes: &ret.numNodes,
+		ExecPath:     execPath,
+		NumNodes:     &ret.numNodes,
+		ChainConfigs: ret.chainConfigs,
```
holisticode, line 112: Is there a reason why we are not asking for this like for the other `StartRequest` parameters and set it right away? I.e.
```
if ret.chainConfigs != nil {
...
}
```

felipemadero: considering the nil case and the not nil case for ret.chainConfigs, just assigning is the same as using an if, but the former is shorter

PR #134, opened 2022-04-27
Harmonize node config hierarchy

holisticode:
closes https://ava-labs.atlassian.net/browse/FRAM-300


cmd/avalanche-network-runner/control/control.go
```diff
@@ -100,10 +101,16 @@ func newStartCommand() *cobra.Command {
...
 	cmd.PersistentFlags().StringVar(
-		&nodeConfig,
-		"node-config",
+		&defaultNodeConfig,
+		"default-node-config",
 		"",
-		"node config as string",
+		"default node config as string, applied to all nodes",
+	)
+	cmd.PersistentFlags().StringVar(
```
danlaine, line 109: We should give a better explanation of this flag's meaning. e.g. What happens if you don't pass it? What are the keys and values in the map the user passes? What if a flag is different than one in the default config?

Also, what would you think of using `StringToStringVar` instead of `StringVar`? i.e. the user does

```
--custom-node-configs node1={CONFIG GOES HERE},node2={CONFIG GOES HERE}
```

might be cleaner

server/network.go
```diff
@@ -171,22 +174,55 @@ func newLocalNetwork(opts localNetworkOptions) (*localNetwork, error) {
 	}, nil
 }
 
-func createConfigFileString(nodeConfig string, logLevel string, logDir string, dbDir string, pluginDir string, whitelistedSubnets string) (string, error) {
-	var jsonContent map[string]interface{}
-	if err := json.Unmarshal([]byte(nodeConfig), &jsonContent); err != nil {
-		return "", err
+// get the node configs right
```
danlaine, line 177: Can we please make this comment a bit more clear as to what it takes in and returns?

server/network.go
```diff
@@ -171,22 +174,55 @@ func newLocalNetwork(opts localNetworkOptions) (*localNetwork, error) {
...
+func mergeNodeConfig(defaultConfig string, customConfig string, globalConfig string) (map[string]interface{}, error) {
```
danlaine, line 178: I think it would make sense to cache the unmsarshalled representations of `defaultConfig` and `globalConfig`  (that is, store them as `map[string]interface{}`) instead of storing them as `string` and then repeatedly unmarshalling them here.

server/server.go
```diff
@@ -549,23 +556,23 @@ func (s *server) AddNode(ctx context.Context, req *rpcpb.AddNodeRequest) (*rpcpb
...
 	}
-	if req.StartRequest.WhitelistedSubnets != nil {
-		whitelistedSubnets = *req.StartRequest.WhitelistedSubnets
-	}
-	if req.StartRequest.PluginDir != nil {
-		pluginDir = *req.StartRequest.PluginDir
-	}
+
+	// use same configs from other nodes
+	whitelistedSubnets = s.network.nodeInfos[s.network.nodeNames[0]].WhitelistedSubnets
```
danlaine, line 561: we should document this behavior

server/server.go
```diff
@@ -549,23 +556,23 @@ func (s *server) AddNode(ctx context.Context, req *rpcpb.AddNodeRequest) (*rpcpb
...
+	pluginDir = s.network.nodeInfos[s.network.nodeNames[0]].PluginDir
```
danlaine, line 562: we should document this behavior

danlaine (changes requested):

cmd/avalanche-network-runner/control/control.go
holisticode: I opted to use `StringVar` to be consistent with the other map in this command, `customVMNameToGenesisPath,`

As for the better explanation - there is only so much which can go into the param description for the command...I'll give it a shot improving it, but tbh I think the extensive explanation should go to the docs.

server/network.go
holisticode: Good suggestion. Fixed.

server/server.go
holisticode: You mean in README/docs right?
Or somewhere else

server/server.go
holisticode: You mean in README/docs right?

cmd/avalanche-network-runner/control/control.go
```diff
@@ -100,10 +101,16 @@ func newStartCommand() *cobra.Command {
...
+		"default-node-config",
 		"",
-		"node config as string",
+		"[optional] default node config as JSON string, applied to all nodes",
+	)
+	cmd.PersistentFlags().StringVar(
+		&customNodeConfigs,
+		"custom-node-configs",
+		"",
+		"[optional] custom node configs as JSON string of map, for each node individually. Common entries verride `default-node-config`, but can be combined. Invalidates `number-of-nodes` (provide all node configs if uses).",
```
felipemadero, line 113: In principle I don't see the need to invalidate number-of-nodes. You can just override some nodes. Or am I missing something?
Regarding the help, a suggestion:
"[optional] custom configs for each node. with precedence over `default-node-config`. invalidates `number-of-nodes` (you must provide configs for all nodes)."

server/network.go
```diff
@@ -171,22 +186,51 @@ func newLocalNetwork(opts localNetworkOptions) (*localNetwork, error) {
...
+// returns final map of node config entries
+func mergeNodeConfig(baseConfig map[string]interface{}, globalConfig map[string]interface{}, customConfig string) (map[string]interface{}, error) {
+	var jsonCustom map[string]interface{}
+
+	// merge, overwriting entries in default with the global ones
+	for k, v := range globalConfig {
+		baseConfig[k] = v
 	}
+
+	if customConfig != "" {
```
felipemadero, line 202: Ok now I see that custom node configs len is not invalidating number of nodes param... That is correct behaviour for me. Shall only change then the description of the cobra command.

server/network.go
```diff
@@ -111,6 +113,18 @@ func newLocalNetwork(opts localNetworkOptions) (*localNetwork, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	var defaultConfig, globalConfig map[string]interface{}
```
felipemadero, line 117: probably the names defaultConfig, globalConfig are misleading (globalConfig is named defaultConfig elsewhere), shall try to preserv defaultConfig related naming for the last one.
suggestion... staticDefaultNodeConfig, userProvidedDefaultNodeConfig, customNodeConfig. Anyway, just a suggestion.

server/network.go
```diff
@@ -120,12 +134,13 @@ func newLocalNetwork(opts localNetworkOptions) (*localNetwork, error) {
 		nodeNames[i] = nodeName
 		cfg.NodeConfigs[i].Name = nodeName
 
-		// get the node configs right
-		nodeConfig := defaultNodeConfig
-		if opts.nodeConfigParam != "" {
-			nodeConfig = opts.nodeConfigParam
+		var mergedConfig map[string]interface{}
+		mergedConfig, err = mergeNodeConfig(defaultConfig, globalConfig, opts.customNodeConfigs[nodeNames[i]])
```
felipemadero, line 138: A probably useful function can be created that merges any slice of maps with the expected precedence.
Custom node config shall be unmarshalled externally in that case.

server/server.go
```diff
@@ -292,13 +294,18 @@ func (s *server) Start(ctx context.Context, req *rpcpb.StartRequest) (*rpcpb.Sta
...
 		zap.String("pluginDir", pluginDir),
-		zap.String("nodeConfig", nodeConfigParam),
+		zap.String("defaultNodeConfig", defaultNodeConfig),
 	)
 
 	if s.network != nil {
 		return nil, ErrAlreadyBootstrapped
 	}
 
+	if len(customNodeConfigs) > 0 {
```
felipemadero, line 304: I don't understand why this is needed. Also shall avoid it (not force the user to given custom configs for all nodes)

server/server.go
```diff
@@ -549,23 +557,33 @@ func (s *server) AddNode(ctx context.Context, req *rpcpb.AddNodeRequest) (*rpcpb
...
 	}
-	if req.StartRequest.WhitelistedSubnets != nil {
-		whitelistedSubnets = *req.StartRequest.WhitelistedSubnets
-	}
-	if req.StartRequest.PluginDir != nil {
-		pluginDir = *req.StartRequest.PluginDir
-	}
+
+	// use same configs from other nodes
+	whitelistedSubnets = s.network.nodeInfos[s.network.nodeNames[0]].WhitelistedSubnets
```
felipemadero, line 562: probably shall add whitelistedSubnets and pluginDir fields to localNetwork struct

server/network.go
holisticode: https://github.com/ava-labs/avalanche-network-runner/pull/134/files/f7eb4b273a525fbdb1333d0a76b33e7f8666ccb5#diff-bcef29333b5dd1d156e40416d98c682d7b3d16d1d7485ca8ad885ba25958d39f

server/network.go
holisticode: Already updated

cmd/avalanche-network-runner/control/control.go
danlaine: `I opted to use StringVar to be consistent with the other map in this command, customVMNameToGenesisPath`

I think it would probably make sense to use `StringToStringVar` there or any time we have a map from string to string. What do you think @gyuho ?

server/server.go
danlaine: and in the code, as necessary

server/server.go
danlaine: and in the code, as necessary

server/network.go
holisticode: Applied `globalConfig` to be consistent everywhere

server/server.go
holisticode: Decided to store the whole options, because those were the actual initialization params.

server/server.go
holisticode: Updated README. Will also add to public docs.

server/server.go
holisticode: Updated in README

cmd/avalanche-network-runner/control/control.go
holisticode: With `StringVar` we keep the option to just use `--flagname $(echo config.json)` , which allows to maintain the config in a file which is easier to edit (especially for bigger configs).

server/network.go
```diff
@@ -171,22 +190,65 @@ func newLocalNetwork(opts localNetworkOptions) (*localNetwork, error) {
...
+// globalConfig: map of global config provided to be applied to all nodes. Overrides defaultConfig
+// customConfig: a custom config provided to be applied to this node. Overrides globalConfig and defaultConfig
+// returns final map of node config entries
+func mergeNodeConfig(baseConfig map[string]interface{}, globalConfig map[string]interface{}, customConfig string) (map[string]interface{}, error) {
+	var jsonCustom map[string]interface{}
+
+	mergeAndCheckForIgnores(baseConfig, globalConfig)
+
+	// merge, overwriting entries in default with the global ones
+	if customConfig != "" {
```
danlaine, line 222:
```suggestion
	mergeAndCheckForIgnores(baseConfig, globalConfig)

	// merge, overwriting entries in default with the global ones
	var jsonCustom map[string]interface{}
	if customConfig != "" {
```

danlaine (changes requested):

danlaine:
All pretty small comments/requests.  Otherwise LGTM!

README.md
```diff
@@ -188,6 +229,47 @@ avalanche-network-runner control restart-node \
...
+# or
+avalanche-network-runner control add-node \
+--request-timeout=3m \
+--log-level debug \
+--endpoint="0.0.0.0:8080" \
+--node-name node99 \
+--avalanchego-path ${AVALANCHEGO_EXEC_PATH}
+```
+
+Adding a node also supports some of the same parameters to start the network resp. to run subnets:
```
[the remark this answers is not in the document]
holisticode: I am wanting to say something like.... the same parameters to start the network (start command) without and also with subnets...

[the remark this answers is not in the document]
danlaine: Maybe something like this?

```
You can also provide additional flags that specify the node's config, and what custom VMs it supports. 
```

cmd/avalanche-network-runner/control/control.go
```diff
@@ -411,20 +416,20 @@ func addNodeFunc(cmd *cobra.Command, args []string) error {
 	}
 	defer cli.Close()
 
-	if nodeConfig != "" {
-		color.Outf("{{yellow}}WARNING: overriding node configs with custom provided config {{/}} %+v\n", nodeConfig)
+	opts := []client.OpOption{
+		client.WithLogLevel(logLevel),
+	}
+
+	if globalNodeConfig != "" {
```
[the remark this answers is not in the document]
holisticode: OK used `addNodeConfig` - which really is just the variable, as the cli param is indeed called `node-config`

README.md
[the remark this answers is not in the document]
holisticode: ok.

README.md
```diff
@@ -109,9 +109,49 @@ curl -X POST -k http://localhost:8081/v1/control/start -d '{"execPath":"'${AVALA
...
+3. `--custom-node-configs` is a map of JSON strings representing the *complete* network with individual configs. This allows to configure each node independently. If set, `--number-of-nodes` will be **ignored** to avoid conflicts.
+4. The configs can be combined and will be merged, i.e. one could set global `--global-node-config` entries applied to each node, and also set `--custom-node-configs` for additional entries.
+5. Common `--custom-node-configs` entries override `--global-node-config` entries which override the standard set.
+6. The following entries will be **ignored in all cases** because the network-runner needs to set them internally to function properly:
+    ```bash
+      --log-dir
+      --db-dir
+      --http-port
+      --staking-port
+      --public-ipc
```
danlaine, line 151:
```suggestion
      --public-ip
```

cmd/avalanche-network-runner/control/control.go
```diff
@@ -100,10 +102,16 @@ func newStartCommand() *cobra.Command {
...
+		"global-node-config",
 		"",
-		"node config as string",
+		"[optional] global node config as JSON string, applied to all nodes",
+	)
+	cmd.PersistentFlags().StringVar(
+		&customNodeConfigs,
+		"custom-node-configs",
+		"",
+		"[optional] custom node configs as JSON string of map, for each node individually. Common entries override `global-node-config`, but can be combined. Invalidates `number-of-nodes` (provide all node configs if uses).",
```
danlaine, line 114:
```suggestion
		"[optional] custom node configs as JSON string of map, for each node individually. Common entries override `global-node-config`, but can be combined. Invalidates `number-of-nodes` (provide all node configs if used).",
```

cmd/avalanche-network-runner/control/control.go
```diff
@@ -411,20 +415,20 @@ func addNodeFunc(cmd *cobra.Command, args []string) error {
...
+
+	if addNodeConfig != "" {
+		color.Outf("{{yellow}}WARNING: overriding node configs with custom provided config {{/}} %+v\n", addNodeConfig)
 		// validate it's valid JSON
 		var js json.RawMessage
-		if err := json.Unmarshal([]byte(nodeConfig), &js); err != nil {
+		if err := json.Unmarshal([]byte(addNodeConfig), &js); err != nil {
 			return fmt.Errorf("failed to validate JSON for provided config file: %s", err)
 		}
+		opts = append(opts, client.WithGlobalNodeConfig(addNodeConfig))
```
danlaine, line 429: Is this supposed to be `WithGlobalNodeConfig`? This isn't a global config, and I don't think we do anything with the `globalNodeConfig` option in `AddNode`

tests/e2e/e2e_test.go
```diff
@@ -200,7 +211,7 @@ var _ = ginkgo.Describe("[Start/Remove/Restart/Add/Stop]", func() {
 
 	ginkgo.It("can wait for health", func() {
 		// start is async, so wait some time for cluster health
-		time.Sleep(2 * time.Minute)
+		time.Sleep(30 * time.Second)
```
danlaine, line 214: Can we just poll every 10 seconds for health instead of sleeping here?

tests/e2e/e2e_test.go
```diff
@@ -317,4 +328,62 @@ var _ = ginkgo.Describe("[Start/Remove/Restart/Add/Stop]", func() {
 			color.Outf("{{green}}add-node failed as expected")
 		})
 	})
+
+	ginkgo.It("can start with custom config", func() {
+		ginkgo.By("stopping network first", func() {
```
danlaine, line 333: Do we need to shut down the existing network for this? Can't we just start a new, separate network?

cmd/avalanche-network-runner/control/control.go
holisticode: Ok so AddNode contains `StartRequest` as its gRPC data container because there are common params. In `StartRequest` we only have `GlobalConfig` though, no individual node-config entry.

On the server side, `AddNode` indeed *does* read `req.StartRequest.GetGlobalNodeConfig()` to then evaluate the config, the same way it does when receiving an actual `StartRequest`.

I don't think that's too bad. However, we could add a `NodeConfig` entry to the `AddNodeRequest` container if you insist, and use that instead also on server side.

tests/e2e/e2e_test.go
holisticode: Can we please do this in a separate PR? This pattern is used all over the e2e tests and it has not been introduced with this PR.

tests/e2e/e2e_test.go
holisticode: We can't. Starting a new network results in an error "already bootstrapped" - at least that was my experience.

README.md
```diff
@@ -109,9 +109,49 @@ curl -X POST -k http://localhost:8081/v1/control/start -d '{"execPath":"'${AVALA
...
+        "api-ipcs-enabled":true,
+        "index-enabled":true
+        }
+    ```
+2. `--global-node-config` is a JSON string representing a *single* avalanchego config, which will be applied to **all nodes**. This makes it easy to define common properties to all nodes. Whatever is set here will be *combined* with the standard set above.
+3. `--custom-node-configs` is a map of JSON strings representing the *complete* network with individual configs. This allows to configure each node independently. If set, `--number-of-nodes` will be **ignored** to avoid conflicts.
+4. The configs can be combined and will be merged, i.e. one could set global `--global-node-config` entries applied to each node, and also set `--custom-node-configs` for additional entries.
+5. Common `--custom-node-configs` entries override `--global-node-config` entries which override the standard set.
+6. The following entries will be **ignored in all cases** because the network-runner needs to set them internally to function properly:
+    ```bash
```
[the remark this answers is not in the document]
holisticode: Not a big deal, but why? All cli code has this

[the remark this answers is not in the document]
danlaine: The text in the snippet isn't bash

tests/e2e/e2e_test.go
holisticode: Agreed. I will also create a ticket right away.

README.md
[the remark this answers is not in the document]
holisticode: Strictly speaking you are right. Will do.

server/network.go
```diff
@@ -120,12 +142,12 @@ func newLocalNetwork(opts localNetworkOptions) (*localNetwork, error) {
 		nodeNames[i] = nodeName
 		cfg.NodeConfigs[i].Name = nodeName
 
-		// get the node configs right
-		nodeConfig := defaultNodeConfig
-		if opts.nodeConfigParam != "" {
-			nodeConfig = opts.nodeConfigParam
+		mergedConfig, err := mergeNodeConfig(defaultConfig, globalConfig, opts.customNodeConfigs[nodeNames[i]])
```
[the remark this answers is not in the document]
danlaine: I think it's kinda broken if the user has to name the node `node[number]`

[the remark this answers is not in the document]
holisticode: You are probably right. Added NOTEs to README and code at several places. Will create a new ticket and address this next.

README.md
```diff
@@ -109,9 +109,49 @@ curl -X POST -k http://localhost:8081/v1/control/start -d '{"execPath":"'${AVALA
...
+	--global-node-config '{"index-enabled":false, "api-admin-enabled":true,"network-peer-list-gossip-frequency":"300ms"}'
+	--custom-node-configs" '{"node1":{"log-level":"debug","api-admin-enabled":false},"node2":{...},...}'
+```
+
+`--plugin-dir` and `--custom-vms` are parameters relevant to subnet operation.
+See the [subnet](#network-runner-rpc-server-subnet-evm-example) section for details about how to run subnets.
+
+The network-runner supports avalanchego node configuration at different levels.
+1. If neither `--global-node-config` nor `--custom-node-configs` is supplied, all nodes get a standard set of config options. Currently this set contains:
+    ```bash
```
danlaine, line 130:
```suggestion
    ```json
```

README.md
danlaine, line 146:
```suggestion
    ```
```

README.md
```diff
@@ -188,6 +228,46 @@ avalanche-network-runner control restart-node \
...
+--request-timeout=3m \
+--log-level debug \
+--endpoint="0.0.0.0:8080" \
+--node-name node99 \
+--avalanchego-path ${AVALANCHEGO_EXEC_PATH}
+```
+
+You can also provide additional flags that specify the node's config, and what custom VMs it supports:
+
+```bash
```
danlaine, line 252:
```suggestion
```
```

**Review by danlaine — COMMENTED** [review]
Ok these are my actual final comments for this PR. @gyuho Would you mind reviewing this PR as well? Especially the new E2E test :pray:

File: tests/e2e/e2e_test.go (hunk 19 above)
Comment (danlaine) [reply tests/e2e/e2e_test.go:334]: To give you a bit more context @gyuho -- what we'd like to do is have an E2E test that makes sure that the `--custom-node-configs` flag works. To do that, we need to start a new network with that flag. What's the best way to do that (start a new network) in the E2E tests?

File: README.md [hunk 24]
```diff
@@ -109,9 +109,50 @@ curl -X POST -k http://localhost:8081/v1/control/start -d '{"execPath":"'${AVALA
...
 --endpoint="0.0.0.0:8080" \
+--number-of-nodes=5 \
 --avalanchego-path ${AVALANCHEGO_EXEC_PATH}
 ```
 
+Additional optional parameters which can be passed to the start command:
+
+```bash
+  --plugin-dir ${AVALANCHEGO_PLUGIN_PATH} \
+	--custom-vms '{"subnetevm":"/tmp/subnet-evm.genesis.json"}'
```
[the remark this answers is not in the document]
gyuho: ref. https://github.com/ava-labs/avalanche-network-runner/blob/main/README.md#network-runner-rpc-server-subnet-evm-example

gyuho:
Can we make sure the example here still works https://github.com/ava-labs/avalanche-network-runner/blob/main/README.md#network-runner-rpc-server-subnet-evm-example?

Otherwise, lgtm. thx!

gyuho (approved):
lgtm thx!

PR #138, opened 2022-05-02
fix CLI usage in README

jacobhjkim:
`--message-bytes-throttling` is not a valid flag.


felipemadero:
> > `--message-bytes-throttling` is not a valid flag.
> 
> thanks for this. updating and merging.

making the change using 
https://github.com/ava-labs/avalanche-network-runner/pull/173 , as this one has unsigned commits

thanks!

PR #140, opened 2022-05-07
rpcpb: remove optional usage

hexfusion:
As of protobuf 3.15[1] `optional` field support was added to proto3 spec aligning with proto2 functionality. In rust this usage results in a panic because it is using a previous protobuf version.
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "protoc failed: rpc.proto:156:12: Explicit 'optional' labels are disallowed in the Proto3 syntax. To define 'optional' fields in Proto3, simply remove the 'optional' label, as fields are 'optional' by default.\nrpc.proto:157:12: Explicit 'optional' labels are disallowed in the Proto3 syntax.
```

This PR removes the optional fields from the rpcpb proto.

[1] https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0


hexfusion:
going to handle this in another way.

hexfusion:
https://github.com/ava-labs/avalanche-network-runner-sdk-rs/pull/3

PR #141, opened 2022-05-09
Fix health call race condition while installing vms

holisticode:
Proposal to fix possible race condition when the network is restarting nodes during a custom VM install and at the same time call `Health` via gRPC.

To be questioned! Might be a symptom rather than the cause.

### DIFF
--- a/local/network.go
+++ b/local/network.go
@@ -436,9 +436,6 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
 
 // See network.Network
 func (ln *localNetwork) Healthy(ctx context.Context) chan error {
-	ln.lock.RLock()
-	defer ln.lock.RUnlock()
-
 	zap.L().Info("checking local network healthiness", zap.Int("nodes", len(ln.nodes)))
 	healthyChan := make(chan error, 1)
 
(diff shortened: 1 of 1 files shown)


local/network.go
```diff
@@ -448,11 +445,15 @@ func (ln *localNetwork) Healthy(ctx context.Context) chan error {
 		return healthyChan
 	}
 
-	nodes := make([]*localNode, 0, len(ln.nodes))
-	for _, node := range ln.nodes {
-		nodes = append(nodes, node)
-	}
 	go func() {
+		ln.lock.RLock()
```
felipemadero, line 449: this works, but it keeps the network locked during the health check. do we want that?
if the answer is no, then a more complex health is needed, that restart if the list of nodes changes.

local/network.go
```diff
@@ -448,11 +445,15 @@ func (ln *localNetwork) Healthy(ctx context.Context) chan error {
...
+		defer ln.lock.RUnlock()
+
+		nodes := make([]*localNode, 0, len(ln.nodes))
```
felipemadero, line 452: in this situation there should be no need to copy ln.nodes to nodes.

felipemadero (commented):
this works, but it keeps the network locked during the health check. do we want that?
if the answer is no, then a more complex health is needed, that restart if the list of nodes changes.

local/network.go
holisticode: Maybe someone else should have a say at this one, but honestly I think it's not too bad to lock during the health check. When a DB is queried you also read-lock otherwise you might get stale results...

local/network.go
holisticode: fixed

local/network.go
holisticode: The problem here is that the health check itself waits, so we should probably fix that - but possibly in another PR

felipemadero: As per solving the current race condition this work. LGTM would approve but let's please add a TODO comment for it, probably mentioning possible future implementation

felipemadero (approved):

PR #146, opened 2022-05-11
add network stop on server kill signal

felipemadero:


server/server.go
```diff
@@ -177,6 +178,14 @@ func (s *server) Run(rootCtx context.Context) (err error) {
 
 		s.gRPCServer.Stop()
 		zap.L().Warn("closed gRPC server")
+
+		if s.network != nil {
+			stopCtx, stopCtxCancel := context.WithTimeout(context.Background(), StopOnSignalTimeout)
+			defer stopCtxCancel()
+			s.network.stop(stopCtx)
+			zap.L().Warn("stoping network")
```
gyuho, line 186: Is `stop` async operation? Should we add warning before we call `stop`?

holisticode: I think it should say "stopping"

holisticode: I think it's not async from here.

felipemadero: stop is blocking. changed msg to "network stopped"

server/server.go
```diff
@@ -177,6 +178,14 @@ func (s *server) Run(rootCtx context.Context) (err error) {
 
 		s.gRPCServer.Stop()
 		zap.L().Warn("closed gRPC server")
+
+		if s.network != nil {
```
holisticode, line 182: Question: why are we applying this only to the `case err = <-gwErrc` case? Shouldn't this be done in all cases?

server/server.go
felipemadero: already changed

server/server.go
felipemadero: we were only applying only to the rootCtx closure (dure to signal handling). But you are right. Moving this to outside the select.

holisticode (approved):

PR #149, opened 2022-05-13
Control nodes log

felipemadero:
- remove --log-level param from RPC API. going to use global-node-config
- add --disable-nodes-output to server cmd to disable all node output on server
- improve cmd args spec to avoid spurius extra arguments on commands


cmd/avalanche-network-runner/control/control.go
```diff
@@ -489,7 +502,13 @@ func restartNodeFunc(cmd *cobra.Command, args []string) error {
 	defer cli.Close()
 
 	ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
-	info, err := cli.RestartNode(ctx, nodeName, client.WithExecPath(avalancheGoBinPath), client.WithWhitelistedSubnets(whitelistedSubnets))
+	info, err := cli.RestartNode(
+		ctx,
+		nodeName,
+		client.WithExecPath(avalancheGoBinPath),
+		client.WithWhitelistedSubnets(whitelistedSubnets),
+		client.WithNodeLogLevel(nodeLogLevel),
```
holisticode, line 510: It's a bit weird seeing only the Node Log level here but no node configs?

cmd/avalanche-network-runner/server/server.go
```diff
@@ -22,25 +22,28 @@ func init() {
...
-	port        string
-	gwPort      string
-	gwDisabled  bool
-	dialTimeout time.Duration
+	logLevel           string
+	port               string
+	gwPort             string
+	gwDisabled         bool
+	dialTimeout        time.Duration
+	disableNodesOutput bool
```
holisticode, line 30: Why do we want to disable node log output?

pkg/logutil/node.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// DefaultNodeLogLevel is the default node log level.
+var DefaultNodeLogLevel = "info"
+
+func AvalanchegoToCorethLogLevel(logLevel string) (string, error) {
```
holisticode, line 15: Why do we need this?

rpcpb/rpc.proto
```diff
@@ -155,7 +155,7 @@ message StartRequest {
   string exec_path                    = 1;
   optional uint32 num_nodes           = 2;
   optional string whitelisted_subnets = 3;
-  optional string log_level           = 4;
+  optional string node_log_level      = 4;
```
holisticode, line 158: Note this might break existing scripts. 

cc @gyuho

server/network.go
```diff
@@ -149,14 +152,26 @@ func newLocalNetwork(opts localNetworkOptions) (*localNetwork, error) {
...
-		cfg.NodeConfigs[i].ConfigFile, err = createConfigFileString(mergedConfig, logLevel, logDir, dbDir, opts.pluginDir, opts.whitelistedSubnets)
+		cfg.NodeConfigs[i].ConfigFile, err = createConfigFileString(mergedConfig, nodeLogLevel, logDir, dbDir, opts.pluginDir, opts.whitelistedSubnets)
+		if err != nil {
+			return nil, err
+		}
+
+		if cfg.NodeConfigs[i].CChainConfigFile == "" {
+			cfg.NodeConfigs[i].CChainConfigFile = "{}"
+		}
+		cchainLogLevel, err := logutil.AvalanchegoToCorethLogLevel(nodeLogLevel)
```
holisticode, line 160: Why are we changing this only item for the cchain config? It feels like a bit arbitrary just only doing this.

Maybe we should consider the whole cchain (and other custom chain configs?) more globally, as a whole?

holisticode (commented):
The idea of making the node log configuration clear is good.

But do we really want to do this this way? It feels becoming kinda complex.
1. Do we want to have the node log level be a separate cli param? Maybe we could just enforce it to be via node-config only, then we could just forget most changes here
2. It's not clear to me why we are addressing the cchain config file only for this. I feel like we should address the cchain config more globally (as a whole)
3. I don't understand either the need for the avalanche mapping levels

cmd/avalanche-network-runner/control/control.go
felipemadero: There was a separate PR for node configs. The treatment of global/individual node configs among start, addnode, restartnode is not so uniform. I think shall be part of a new, different PR.

cmd/avalanche-network-runner/server/server.go
felipemadero: This is a fast server option. Makes fast filtering of logs using process control output and not avalanchego/pluging configs. Think it is useful for code iteration, to just avoid msgs and be able to fast look at the one you want.

pkg/logutil/node.go
felipemadero: The server received an avalanchego log level but shall give a go-ethereum log level to coreth. Without also controlling coreth logs, there is no complete control of log level. Not considering other plugins for the moment.

server/network.go
felipemadero: The PR is trying to manage node logs considering only avalanchego+coreth. Usually the common case for the moment. You need to pass cchain config to coreth to get its log level modified. Probably be need another PR for changing all the other cchain config using flags. For other plugins, this seems out of scope, refers to avalanchego pluging stdout/stderr log management/filtering.

felipemadero:
> The idea of making the node log configuration clear is good.
> 
> But do we really want to do this this way? It feels becoming kinda complex.
> 
> 1. Do we want to have the node log level be a separate cli param? Maybe we could just enforce it to be via node-config only, then we could just forget most changes here

It existed previously as log-level flag but was not working for certain avalanchego levels, like "off". In this sense, this is a fix for that.

> 2. It's not clear to me why we are addressing the cchain config file only for this. I feel like we should address the cchain config more globally (as a whole)

Agree. But it another PR.

> 3. I don't understand either the need for the avalanche mapping levels

Needed to pass only one level specification and get is passed to both avalanchego and coreth.

cmd/avalanche-network-runner/control/control.go
holisticode: OK. Please create a ticket for that.

pkg/logutil/node.go
holisticode: I don't think this should be here. If it's needed, it should go into avalanchego, otherwise every library using it has to do it.
Also, I have never seen that we config used for avalanchego is applied automagically to the cchain...That's why different configs exists.

I don't think  we should do this....

server/network.go
holisticode: https://github.com/ava-labs/avalanche-network-runner/pull/149/files#r872665682

holisticode (approved):

hexfusion:
Breaking changes to RPCs should be called out in the CHANGELOG. Added a label.

PR #150, opened 2022-05-13
Add node process failure management

felipemadero:


local/node.go
```diff
@@ -38,27 +38,50 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
+	cmd              *exec.Cmd
+	running          bool
+	waitReturnCh     chan error
+	unexpectedStopCh chan string
 }
 
-func (p *nodeProcessImpl) Start() error {
-	return p.cmd.Start()
+func (p *nodeProcessImpl) Start(name string, unexpectedStopCh chan string) error {
+	p.unexpectedStopCh = unexpectedStopCh
```
danlaine, line 58: 1. I think `node` should be part of the `nodeProcessImpl` state rather than an argument -- we already have the name when we create this struct in  `NewNodeProcess`
2.  I think the user should pass in a `chan int` rather than `chan string`, and we should send the exit code (`p.cmd.ProcessState.ExitCode()`) rather than the node name. 
3. I think we should change the naming from `unexpectedStopCh` to just `onStopChan`. The user can decide for themselves whether the stop was expected or not. If the user gets something on this channel, and they didn't call `Stop()`, they know it's unexpected.

local/node.go
```diff
@@ -38,27 +38,50 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
+			p.running = false
+			p.unexpectedStopCh <- name
+		}
+	}()
+	return startErr
 }
 
 func (p *nodeProcessImpl) Wait() error {
-	return p.cmd.Wait()
+	return <-p.waitReturnCh
```
danlaine, line 74: If we make the changes I suggest for `Start()` (namely passing in an `onStopChan`) we don't really need this method. To wait for the node to stop, you'd just wait on that channel.

If we do keep this, it should return an error if `Start()` hasn't been called. I also think it might make sense for it to return `nil` immediately  if `Stop()` has already been called.

Right now, if you call wait twice, the second call will block forever. If we keep this method, we should close `p.waitReturnCh` in the goroutine in `Start`

local/node.go
```diff
@@ -38,27 +38,50 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
 }
 
 func (p *nodeProcessImpl) Stop() error {
```
danlaine, line 77: It might be nice for the user to be able to pass in a context, and if the context is cancelled but the process hasn't yet exited, we send a sigkill to ensure cleanup.

local/node.go
```diff
@@ -193,3 +216,7 @@ func (node *localNode) GetP2PPort() uint16 {
 func (node *localNode) GetAPIPort() uint16 {
 	return node.apiPort
 }
+
+func (node *localNode) GetAliveStatus() bool {
```
danlaine, line 220:
```suggestion
func (node *localNode) Alive() bool {
```

local/node.go
danlaine: I think this should be guarded by a `sync.Once` to make sure it only runs once. Idk what happens when you signal a dead process and it doesn't seem to be well documented.

local/node.go
danlaine: I also think we should guard this with a `sync.Once` to ensure it only runs once

local/node.go
```diff
@@ -38,27 +38,50 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
+	running          bool
+	waitReturnCh     chan error
+	unexpectedStopCh chan string
 }
 
-func (p *nodeProcessImpl) Start() error {
-	return p.cmd.Start()
+func (p *nodeProcessImpl) Start(name string, unexpectedStopCh chan string) error {
+	p.unexpectedStopCh = unexpectedStopCh
+	startErr := p.cmd.Start()
```
danlaine, line 59: if `startErr != nil` we should immediately return an error

network/node/node.go
```diff
@@ -34,6 +34,8 @@ type Node interface {
 	// It's left to the caller to maintain a reference to the returned peer.
 	// The caller should call StartClose() on the peer when they're done with it.
 	AttachPeer(ctx context.Context, handler router.InboundHandler) (peer.Peer, error)
+	// Return true if the node process is executing
+	GetAliveStatus() bool
```
danlaine, line 38:
```suggestion
	Alive() bool
```

server/server.go
```diff
@@ -875,6 +881,29 @@ func (s *server) getClusterInfo() *rpcpb.ClusterInfo {
 	return info
 }
 
+func (s *server) handleUnexpectedNodeStop() {
```
danlaine, line 884: This should return an error rather than panicking

server/server.go
```diff
@@ -875,6 +881,29 @@ func (s *server) getClusterInfo() *rpcpb.ClusterInfo {
...
+	for {
+		nodeName := <-unexpectedNodeStopCh
+		zap.L().Info("received unexpected node stop message", zap.String("node-name", nodeName))
+		s.mu.Lock()
+		if s.clusterInfo == nil {
+			s.mu.Unlock()
+			return
+		}
+		_, ok := s.network.nodeInfos[nodeName]
+		if ok {
```
danlaine, line 898: Is `ok` ever `false`? I didn't look into it but my guess is it's not supposed to ever be `false`, and if that's the case we should log a warning that something that wasn't supposed to happen, did

local/node.go
```diff
@@ -38,27 +38,50 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
 // AvalancheGo binaries in tests
 type NodeProcess interface {
```
danlaine, line 39: We should document the behavior better...namely:

1. What happens if you call Start/Stop multiple times? Or out of order?
2. What happens if you call Wait and Stop had already been called?
3. What happens if you call Alive but Start hasn't been called?

local/network.go
```diff
@@ -458,14 +461,22 @@ func (ln *localNetwork) Healthy(ctx context.Context) chan error {
 				// Every constants.HealthCheckInterval, query node for health status.
 				// Do this until ctx timeout
 				for {
+					// fast response if died previously
+					if !node.GetAliveStatus() {
+						return fmt.Errorf("unexpected stop on node %q", node.GetName())
```
danlaine, line 466:
```suggestion
						return fmt.Errorf("node %q stopped", node.GetName())
```

local/network.go
```diff
@@ -593,16 +604,23 @@ func (ln *localNetwork) removeNode(nodeName string) error {
 	delete(ln.nodes, nodeName)
 	// cchain eth api uses a websocket connection and must be closed before stopping the node,
 	// to avoid errors logs at client
-	node.client.CChainEthAPI().Close()
-	if err := node.process.Stop(); err != nil {
-		return fmt.Errorf("error sending SIGTERM to node %s: %w", nodeName, err)
-	}
+	node.GetAPIClient().CChainEthAPI().Close()
+	// Ctrl+C on terminal causes a kill for all process group
+	// so sometimes node is already stopped here
```
danlaine, line 609: I don't this sending a SIGTERM to the ANR sends a SIGTERM to the children (i.e. avalanchego). That's why we have the signal handling logic

examples/local/indepth/main.go
```diff
@@ -162,6 +176,11 @@ func run(log logging.Logger, binaryPath string) error {
 	log.Info("updated network's nodes: %s", nodeNames)
 	log.Info("Network will run until you CTRL + C to exit...")
 	// Wait until done shutting down network after SIGINT/SIGTERM
-	<-closedOnShutdownCh
-	return nil
+	select {
+	case <-unexpectedNodeStopCh:
+		// Ctrl+C also kills childs before ANR does, so only use the event to stop, no notification
```
danlaine, line 181: Is this true?

local/network.go
```diff
@@ -226,15 +228,16 @@ func newNetwork(
...
+		networkID:            networkID,
+		genesis:              []byte(networkConfig.Genesis),
+		nodes:                map[string]*localNode{},
+		closedOnStopCh:       make(chan struct{}),
+		log:                  log,
+		bootstraps:           beacon.NewSet(),
+		newAPIClientF:        newAPIClientF,
+		nodeProcessCreator:   nodeProcessCreator,
+		flags:                networkConfig.Flags,
+		unexpectedNodeStopCh: make(chan string),
```
danlaine, line 240: This channel isn't buffered, so if someone isn't reading off this channel we're going to block a bunch of goroutines (namely those started in `Start()`) from exiting

danlaine (changes requested):

local/network.go
felipemadero: If the network is not stopped (in which case shall return earlier), then this node is stopped but was not requested to do that. That's why I use "unexpected stop" here and in other places.

local/node.go
felipemadero: - regarding the Once, I added a state to the nodeProcessImpl so all functions known when they shall be called and how many times

local/node.go
```diff
@@ -38,27 +38,50 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
 	Stop() error
 	// Returns when the process finishes exiting
 	Wait() error
+	// Returns if the process is executing
+	Alive() bool
 }
 
 type nodeProcessImpl struct {
-	cmd *exec.Cmd
+	cmd              *exec.Cmd
```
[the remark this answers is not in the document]
felipemadero: added

local/node.go
felipemadero: changed implementation in different direction but addressing this issues. I think node Wait() is needed because onUnexpectedStopChan is used by ANR local lib user, not by ANR local lib, which indeed uses Wait

felipemadero: Probably I shall explain that the channel is meant to be provided to ANR local lib user (the server, or another app). The same channel is passed to all node processes and also passed to ANR local lib user,  so the name is used to identify which node is passing the message. ANR per se does not use the channel.

local/node.go
felipemadero: add context to Wait, and kill with SIGKILL process and descendants if context is cancelled

felipemadero: ensured wait is executed once, and only called from certain node states. 
if you signal a dead linux process with syscall kill, you get int error for that https://man7.org/linux/man-pages/man2/kill.2.html 
I suppose that is propagated in https://pkg.go.dev/syscall#Kill but, ok, it is not documented how they mapped that to error type

server/server.go
felipemadero: it is not supposed to happen but this are different objects maintained with separate logic so it is a fail safe

local/network.go
felipemadero: I'm not finding good documentation on that, but Bash at least by default sends SIGINT to the hole process group on Ctrl+C event. That is a very different behaviour than sending SIGINT signal to the process by using `kill -SIGINT pid`
We have signal handling logic to properly release resources, and propagate signals, but that bash case is special.

felipemadero:
```
package main

import (
    "fmt"
    "os"
    "os/signal"
    "os/exec"
    "syscall"
)

func main() {
    exec.Command("sleep", "100").Start()
	signalChan := make(chan os.Signal, 1)
	signal.Notify(signalChan, syscall.SIGINT)
	sig := <-signalChan
	fmt.Printf("got OS signal %s\n", sig)
}
```
Try to interrupt this with `Ctrl+C` and with `kill -SIGINT` and check the behaviour and if sleep process continue executing.

examples/local/indepth/main.go
felipemadero: refer to previous comment/example

local/network.go
felipemadero: that's true. Added UnexpectedNodeStopMsgBuffSize to prevent that for a large number of msgs.

local/network.go
danlaine: It's possible that we check whether the network is stopped, and it's not, and then we start the outer goroutine, and then the network is stopped. So actually, it's possible that the node isn't alive because the network was stopped.

local/network.go
```diff
@@ -593,16 +605,23 @@ func (ln *localNetwork) removeNode(nodeName string) error {
...
+	_ = node.process.Stop(ctx)
 	if err := node.process.Wait(); err != nil {
-		return fmt.Errorf("node %q stopped with error: %w", nodeName, err)
+		return fmt.Errorf("node %q stopped with wait error: %w", nodeName, err)
 	}
 	return nil
 }
 
+func (ln *localNetwork) GetUnexpectedNodeStopChannel() (chan network.UnexpectedNodeStopMsg, error) {
+	if ln.isStopped() {
```
danlaine, line 619: We have to hold the lock when we call `ln.isStopped`

local/node.go
```diff
@@ -38,27 +42,126 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
 // AvalancheGo binaries in tests
 type NodeProcess interface {
 	// Start this process
-	Start() error
+	// Returns error if not called after instantiation
```
danlaine, line 45: Nit: Can we put periods in these sentences? Otherwise, when you hover over the method in an IDE it's hard to read.

local/node.go
```diff
@@ -38,27 +42,126 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
+		p.state = Stopped
+		close(p.closeOnStop)
+		p.lock.Unlock()
+		if state != Stopping {
+			p.unexpectedStopCh <- network.UnexpectedNodeStopMsg{
+				Name:     p.name,
+				ExitCode: status.ExitStatus(),
+			}
+		}
+	}()
```
danlaine, line 116: Why don't we just use `p.cmd.Wait()` to wait for the program to exit?

local/node.go
```diff
@@ -38,27 +42,126 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
+	if p.state != Initial {
+		return errors.New("start called on invalid state")
+	}
+	p.unexpectedStopCh = unexpectedStopCh
+	if err := p.cmd.Start(); err != nil {
+		return err
+	}
+	p.state = Started
+	p.waitReturnCh = make(chan error, 1)
+	p.closeOnStop = make(chan struct{})
```
danlaine, line 98: We should add a function `newNodeProcessImpl` and initialize these fields in there

local/node.go
```diff
@@ -38,27 +42,126 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
 }
 
-func (p *nodeProcessImpl) Start() error {
-	return p.cmd.Start()
+const (
+	Initial = iota
+	Started
+	Stopping
+	Stopped
+	Waited
```
danlaine, line 82: What does state `Waited` mean? We set the state to `Waited` after `Wait` returns, but we also set the state to `Stopped` in the goroutine inside `Start`, so it it's possible for the state to be either `Waited` or `Stopped` after the node process stops.

local/node.go
danlaine: In general I'd prefer we avoid using raw syscalls unless absolutely necessary

local/node.go
danlaine: And I think that we could do with just one of `waitReturnCh` and `closeOnStop` since they are used pretty much the same (signalled when the process stops)

local/node.go
```diff
@@ -38,27 +42,126 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
+	name string
+	lock sync.RWMutex
+	cmd  *exec.Cmd
+	// to notify Wait() on process stop, and give wait return
+	waitReturnCh chan error
+	// to notify user of not asked process stops
+	unexpectedStopCh chan network.UnexpectedNodeStopMsg
+	// maintains process state Initial/Started/Stopping/Stopped/Waited
+	state int
+	// to notify SIGKILL goroutine of process end
```
danlaine, line 73:
```suggestion
	// Closed when the AvalancheGo process returns.
```

local/node.go
```diff
@@ -38,27 +42,126 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
 
 type nodeProcessImpl struct {
-	cmd *exec.Cmd
+	name string
+	lock sync.RWMutex
+	cmd  *exec.Cmd
+	// to notify Wait() on process stop, and give wait return
+	waitReturnCh chan error
+	// to notify user of not asked process stops
+	unexpectedStopCh chan network.UnexpectedNodeStopMsg
```
danlaine, line 70: As this code is currently written, we only use `unexpectedStopChan` in `Start`, where we pass it in, so it doesn't need to be a field of this struct

network/network.go
```diff
@@ -39,4 +44,7 @@ type Network interface {
 	// Returns the names of all nodes in this network.
 	// Returns ErrStopped if Stop() was previously called.
 	GetNodeNames() ([]string, error)
+	// Returns a channel that receives a message with
+	// node name when a node fails
+	GetUnexpectedNodeStopChannel() (chan UnexpectedNodeStopMsg, error)
```
danlaine, line 49: We should document that the user should continuously read from this channel, or else it might fill and cause a goroutine leak

local/node.go
```diff
@@ -38,27 +42,126 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
+	waitReturn := <-p.waitReturnCh
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	p.state = Waited
+	return waitReturn
+}
+
+// if context is cancelled, assumes a failure in termination
+// and uses SIGKILL over process and descendants
+func (p *nodeProcessImpl) Stop(ctx context.Context) error {
```
danlaine, line 139: Should `Stop` block until the process is actually stopped?

local/network.go
holisticode: This might be related: https://stackoverflow.com/a/6110213/169252

> The SIGINT signal is generated by the terminal line discipline, and broadcast to all processes in the terminal's _foreground process group_.

local/node.go
holisticode: I agree, this looks like overly complex, we should avoid doing this low-level stuff.

local/node.go
```diff
@@ -38,27 +42,126 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
+	p.lock.RLock()
+	state := p.state
+	p.lock.RUnlock()
+	if state == Initial {
+		return errors.New("wait called on invalid state")
+	}
+	if state == Waited {
+		return nil
+	}
+	waitReturn := <-p.waitReturnCh
```
holisticode, line 130: These blocking channels can cause deadlocks very difficult to debug.

local/node.go
```diff
@@ -38,27 +42,126 @@ type getConnFunc func(context.Context, node.Node) (net.Conn, error)
...
+	}()
+	return nil
 }
 
 func (p *nodeProcessImpl) Wait() error {
-	return p.cmd.Wait()
+	p.lock.RLock()
+	state := p.state
+	p.lock.RUnlock()
+	if state == Initial {
```
holisticode, line 124: Is `Wait` on a state `Stopping` or `Stopped` valid?

local/node.go
```diff
@@ -193,3 +296,28 @@ func (node *localNode) GetP2PPort() uint16 {
...
+		if err != nil {
+			return err
+		}
+		if ppid != pid {
+			continue
+		}
+		if err := killDescendants(proc.Pid); err != nil {
+			return err
+		}
+		_ = syscall.Kill(int(proc.Pid), syscall.SIGKILL)
```
holisticode, line 320: Why not call `proc.Kill` instead of using `syscall`?

rpcpb/rpc.proto
```diff
@@ -141,6 +141,7 @@ message NodeInfo {
   string plugin_dir           = 7;
   string whitelisted_subnets  = 8;
   bytes config                = 9;
+  bool alive                  = 10;
```
holisticode, line 144: Do we really want this? As this is over gRPC, this can generate tricky situations because the client receives `Alive` when the state can be quickly changing on the actual process.

I feel like this is introducing potentially dangerous stuff.

server/server.go
```diff
@@ -875,6 +887,33 @@ func (s *server) getClusterInfo() *rpcpb.ClusterInfo {
 	return info
 }
 
+func (s *server) handleUnexpectedNodeStop(unexpectedNodeStopCh chan network.UnexpectedNodeStopMsg) {
+	for {
+		unexpectedStopMsg := <-unexpectedNodeStopCh
```
holisticode, line 892: Isn't this possibly blocking forever?

network/network.go
```diff
@@ -9,6 +9,11 @@ import (
 
 var ErrStopped = errors.New("network stopped")
 
+type UnexpectedNodeStopMsg struct {
+	Name     string
```
holisticode, line 13: I suggest that `Name` should be changed, as it is confusing about what this `Name` is. If it is always a node name, then `NodeName` maybe.

holisticode (commented):
Pretty impressive what's going on in this PR.

Nevertheless, to me it feels like it's introducing a lot of complexity, and I am not sure I understand what we gain with it. For example I don't understand the need for all the `unexpectedStop` stuff. What is the use case? Can you elaborate on that?

There is also a lot of low level syscalls which maybe can be avoided. It feels kinda dangerous, at least to me. If feels like this can introduce complexity which later down the road can lead to difficult to debug situations.

Just my opinion.

local/network.go
felipemadero: Yes, the idea you illustrate with your example it is the same idea I wanted to clarify with my example. Ctrl+C kills all processes but `kill -SIGINT pid` only kills parent process.

felipemadero: Going back to returning stop errors and improving examples code

felipemadero: I think that when a process that uses ANR receives a SIGINT, it shall asume that certain nodes may die before beign signalled to do so. That is sometimes indeed the case. It shall not asume that all the nodes will be still alive.
So, depending how long the user takes to call network Stop(), the signal sent to processes will found them already dead, and this part of the code will return with error.

examples/local/indepth/main.go
felipemadero: The logic of this PR is that if a process dies before ANR signals it to die, it is considered an unexpected stop.
It does not matter if it dies by having received a signal, for having incorrect params, etc.
When Ctrl+C happens, all processes are signalled before ANR signals them. So it can happen that a process dies before ANR signals it. In relation to the examples, they shall assume that may get unexpected node failure messages after SIGINT (and, really, at this moment you just want to stop the network). Probably a function can be added to ANR to be notified of beign under SIGINT, so as to quickly cancel sending messages of this kind.
I modified the examples so they are faster in communicating and propagating the SIGINT event. 
I think the gain is more by having ANR provide the user this channel for all interesting cases, and ignore the channel for this pretty special case where the shell sends kill to all processes.

felipemadero: The previous example answer the question related to the behaviour of Ctrl+C
Regardong the ignored unexpected stop, I modified the example

local/node.go
felipemadero: Really can use Wait, but there is a race condition in a unit test. Wait closes process pipes and invalidates a unit test case. TestChildCmdRedirection see comment on StdoutPipe, StderrPipe. To avoid that just use Wait4. It seems to me at this point that or the test case is removed or Wait4 is retained.

local/node.go
felipemadero: removed state Waited

local/network.go
```diff
@@ -458,14 +462,22 @@ func (ln *localNetwork) Healthy(ctx context.Context) chan error {
 				// Every constants.HealthCheckInterval, query node for health status.
 				// Do this until ctx timeout
 				for {
+					// fast response if died previously
+					if !node.Alive() {
+						return fmt.Errorf("unexpected stop on node %q", node.GetName())
```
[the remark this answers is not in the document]
felipemadero: which may be the error message? it is like saying, unmanaged stop on node. a node stopped but the ANR did not asked it to stop.

local/node.go
felipemadero: only left one channel. probably easier to follow now.

local/node.go
felipemadero: Wait can be called on a Started, Stopping, Stopped node. In principle it only was to be valid on a Stopping status (Stop() was previously called) or a Stopped status (a notification on stopped process was received, which may happen after a Stop() was called). But after seeing some examples, let the user also call Wait after starting the node, if he prefers, and stop it somewhere else.

rpcpb/rpc.proto
felipemadero: I think it is ok to notify the user on dead node process and let him dedice what to do. Really a process starts alive and is set to not being alive when received the channel notification

server/server.go
felipemadero: here, only the ANR server will read the channel

felipemadero:
> Pretty impressive what's going on in this PR.
> 
> Nevertheless, to me it feels like it's introducing a lot of complexity, and I am not sure I understand what we gain with it. For example I don't understand the need for all the `unexpectedStop` stuff. What is the use case? Can you elaborate on that?
> 
> There is also a lot of low level syscalls which maybe can be avoided. It feels kinda dangerous, at least to me. If feels like this can introduce complexity which later down the road can lead to difficult to debug situations.
> 
> Just my opinion.

At first it was trying to solve an issue with current ANR, that is, when called executed concurrently with duplicated network info, like launching two five node network examples in different terminals, the last execution has all avalanchego processes died, but ANR anyway thinks all is OK as healthy is communicating with the other network. So added already available os process information to avoid node aliases of this kind (or future ones?). A second option to check the status of the nodes in general, and a very different path from using the client API.
Of course that can be avoided in ANR first checks for port availability, but that is somewhat subject to races. The idea is to reduce the burden of the user of the ANR/CLI if he is starting to get, during the usage of the tools, failed networks, networks in the background, or repetition in execution of networks.
 
Second case is for health check to be responsive in the event of a node process beign dead, instead of waiting for health api call success (with timeout).

Third case is for giving the user a way to get fast events and probable had option to manage the situation, when a node dies

local/node_process.go
```diff
@@ -0,0 +1,169 @@
...
+
+	return p.cmd.ProcessState.ExitCode()
+
+}
+
+func (p *nodeProcess) Status() status.Status {
+	if p.firstStatusCall {
+		time.Sleep(firstStatusCallWait)
+		p.firstStatusCall = false
+	}
```
danlaine, line 141: Why do we do this? Also, if we keep this, we should hold the lock here

felipemadero: Sometimes a process will almost inmediatelly fail (say it takes a little less than firstStatusCallWait). For example when trying to use a busy port. A Status() call between calling start() and firstStatusCallWait, will return Running in that case, and a call to Healthy() will also probably return nil (if there is port aliasing). I just try to wait some time before the first Status(), to give the process the chance to almost inmediately fail, and then avoid that situation. (eg, executing twice the example fivenodenetwork)
I dont want to take the lock because I want the status to be updated. Probably I can use a new lock. Really those variables are only used by Status(), so a lock should only be needed for concurrent calls to that function.

local/network.go
```diff
@@ -563,18 +566,24 @@ func (ln *localNetwork) Healthy(ctx context.Context) error {
...
 	for _, node := range ln.nodes {
 		node := node
+		nodeName := node.GetName()
 		errGr.Go(func() error {
 			// Every [healthCheckFreq], query node for health status.
 			// Do this until ctx timeout or network closed.
 			for {
+				if node.Status() != status.Running {
+					// If we had stopped this node ourselves, it wouldn't be in [ln.nodes].
+					// Since it is, it means the node stopped unexpectedly.
```
holisticode, line 576: And should we remove it from `ln.nodes` now or is it not necessary?

local/node_process.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return p.cmd.ProcessState.ExitCode()
+
+}
+
+func (p *nodeProcess) Status() status.Status {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
+
+	if p.firstStatusCall {
```
holisticode, line 141: Is it deterministic when the first status call will succeed? Can we do something else than just sleep, e.g. a channel?

holisticode (commented):
To me this is basically ready, have just 2 small questions left.

local/network.go
felipemadero: Currently it seems not necessary, we are doing the less possible recovery, as it is not clear which action may make the network stable again. A user shall decide what to do based on he is observing. He may call RemoveNode if he finds out that the network can recover from the problem.

local/node_process.go
holisticode: I hoped there would be a way to tell that the network is up and then it's safe to call Status, instead of having to Sleep

felipemadero: I would like also @danlaine opinion on using this wait to check for node

danlaine (approved):

PR #151, opened 2022-05-14
Release lock sooner in `Healthy`

danlaine:


local/network.go
```diff
@@ -440,22 +440,25 @@ func (ln *localNetwork) Healthy(ctx context.Context) chan error {
 	healthyChan := make(chan error, 1)
 
 	// Return unhealthy if the network is stopped
+	ln.lock.RLock()
```
danlaine, line 443: We were supposed to be holding the lock during the call to `isStopped`

local/network.go
```diff
@@ -440,22 +440,25 @@ func (ln *localNetwork) Healthy(ctx context.Context) chan error {
...
+		ln.lock.RUnlock()
 		return healthyChan
 	}
 
-	go func() {
-		// TODO: This will block the network for the duration of the health call.
-		// Maybe a better solution can be found.
-		ln.lock.RLock()
-		defer ln.lock.RUnlock()
+	nodes := make([]*localNode, 0, len(ln.nodes))
```
felipemadero, line 450: a recent fix was related to the lock beign inside the go func(). when:
1) we get a list of nodes
2) the list is modified with Remove
3) the health check continues
In that case the client timeouts and the health check fails

We decided to avoid that failure by (for the moment) forcing the health check to block the add/remove nodes calls, avoiding the changes in the node list

danlaine: I feel like we should just add to the specification of `Healthy` that it gets the health of the nodes that exist at the time `Healthy` is called, and if the user removes a node before `Healthy` has returned, that's their problem

felipemadero: You mean before the channel returned from Healthy receives a message. What do you think on this @holisticode

felipemadero: The problem was that when creating a network with subnets, in that process, the network is being changed in the ANR side, and the ANR also can receive a Healthy call during that and fail

felipemadero: 1) I think @holisticode had the problem of first asking the server to create a network with subnets, and then waiting for that network to be ready. Previously one could just create a network and wait on the health channel. But now health can fail during complex network creation, so something else is needed. I think that for this, maybe a new server function is needed, because the server knows pretty well when if network with subnets has started, or, otherwise, which was the error.
2) Besides that, in local ANR, if RemoveNode shall not be called during Health check, why not lock the network during that check and reduce the user chance to do the wrong thing?

danlaine: I guess it's not a huge deal if we block other operations (i.e. adding or removing a node) while the health check is going on, but I don't love it, either. Something we could do is have a background goroutine that polls nodes for their health every X seconds, and then just report the results of that in `Healthy` to avoid having to block to query the nodes.

felipemadero: I agree maybe we can leave the lock for the moment. We agreed also with Fabio on making this function smarter regarding network changes so as to avoid a lock. Probably another PR. I was thinking in Healthy adapting to changing network conditions before failure. You proposal is also good one to follow.

holisticode: For `avalanche-cli` the issue is that the node is removed and re-added when installing subnets, resulting in the poll for Health to hang forever because it's polling the wrong endpoints. 

It is maybe not very elegant to have to block the network during the health check, but on the other hand, our semantics kind of require that - a network is not operational while it is unhealthy, and if callers depend on that, we can't just presume it is when it is not.

In the traditional DB web2 environment, if you depend on a value for an operation, you also need to read-lock that so that it does not change under your feet. Our health check is more like a consistency check of that DB, not the health check of the web server is running.

Still, the current solution might not be optimal, but it needs deeper thinking and possibly a separate PR. `avalanche-cli` currently can't work without a solution which enforces this.

PR #154, opened 2022-05-19
Make healthy block

danlaine:
* Make `network.Healthy` block and return `error` rather than `chan error`
* Make it so that if `network.Stop` is called while a call to `network.Healthy` is ongoing, `network.Healthy` immediately returns an error

### DIFF
--- a/README.md
+++ b/README.md
@@ -655,16 +655,10 @@ The network runner allows users to interact with an AvalancheGo network using th
 ```go
 // Network is an abstraction of an Avalanche network
 type Network interface {
-  // Returns a chan that is closed when
-  // all the nodes in the network are healthy.
-  // If an error is sent on this channel, at least 1
-  // node didn't become healthy before the timeout.
-  // If an error isn't sent on the channel before it
-  // closes, all the nodes are healthy.
+  // Returns true if all the nodes in the network are healthy.
   // A stopped network is considered unhealthy.
   // Timeout is given by the context parameter.
-  // [ctx] must eventually be cancelled -- if it isn't, a goroutine is leaked.
-  Healthy(context.Context) chan error
+  Healthy(context.Context) error
   // Stop all the nodes.
   // Returns ErrStopped if Stop() was previously called.
   Stop(context.Context) error
--- a/README.md
+++ b/README.md
@@ -655,7 +655,7 @@ The network runner allows users to interact with an AvalancheGo network using th
 ```go
 // Network is an abstraction of an Avalanche network
 type Network interface {
-  // Returns true if all the nodes in the network are healthy.
+  // Returns nil if all the nodes in the network are healthy.
   // A stopped network is considered unhealthy.
   // Timeout is given by the context parameter.
   Healthy(context.Context) error
--- a/local/network.go
+++ b/local/network.go
@@ -463,7 +463,7 @@ func (ln *localNetwork) Healthy(ctx context.Context) error {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 	go func() {
-		// This goroutine runs until Stop is called
+		// This goroutine runs until [ln.Stop] is called
 		// or this function returns.
 		select {
 		case <-ln.onStopCtx.Done():
@@ -479,17 +479,15 @@ func (ln *localNetwork) Healthy(ctx context.Context) error {
 			// Every [healthCheckFreq], query node for health status.
 			// Do this until ctx timeout or network closed.
 			for {
-				select {
-				case <-ln.onStopCtx.Done():
-					return network.ErrStopped
-				case <-cctx.Done():
-					return fmt.Errorf("node %q failed to become healthy within timeout", node.GetName())
-				case <-time.After(healthCheckFreq):
-				}
 				health, err := node.client.HealthAPI().Health(cctx)
 				if err == nil && health.Healthy {
 					ln.log.Debug("node %q became healthy", node.name)
 					return nil
+				}
+				select {
+				case <-cctx.Done():
+					return fmt.Errorf("node %q failed to become healthy within timeout", node.GetName())
+				case <-time.After(healthCheckFreq):
 				}
 			}
 		})
--- a/local/network.go
+++ b/local/network.go
@@ -66,8 +66,18 @@ type localNetwork struct {
 	newAPIClientF api.NewAPIClientF
 	// Used to create new node processes
 	nodeProcessCreator NodeProcessCreator
-	// Closed when network is done shutting down
-	closedOnStopCh chan struct{}
+	stopOnce           sync.Once
+	// Cancelled when Stop begins.
+	onStopCtx context.Context
+	// Cancels [onStopCtx]
+	onStopCtxCancel func()
+	// True if Stop has returned.
+	// This field only exists so that in removeNode,
+	// we know whether Stop has already completed, in
+	// which case we return an error, or not, in which
+	// case we proceed to remove all nodes.
+	// [lock] must be held while accessing this field.
+	stopped bool
 	// For node name generation
 	nextNodeSuffix uint64
 	// Node Name --> Node
@@ -166,7 +176,7 @@ func (npc *nodeProcessCreator) NewNodeProcess(config node.Config, args ...string
 	if config.RedirectStdout {
 		stdout, err := cmd.StdoutPipe()
 		if err != nil {
-			return nil, fmt.Errorf("Could not create stdout pipe: %s", err)
+			return nil, fmt.Errorf("couldn't create stdout pipe: %s", err)
 		}
 		// redirect stdout and assign a color to the text
 		utils.ColorAndPrepend(stdout, npc.stdout, config.Name, color)
@@ -174,7 +184,7 @@ func (npc *nodeProcessCreator) NewNodeProcess(config node.Config, args ...string
 	if config.RedirectStderr {
 		stderr, err := cmd.StderrPipe()
 		if err != nil {
-			return nil, fmt.Errorf("Could not create stderr pipe: %s", err)
+			return nil, fmt.Errorf("couldn't create stderr pipe: %s", err)
 		}
 		// redirect stderr and assign a color to the text
 		utils.ColorAndPrepend(stderr, npc.stderr, config.Name, color)
@@ -224,12 +234,14 @@ func newNetwork(
 		return nil, fmt.Errorf("couldn't get network ID from genesis: %w", err)
 	}
 
+	onStopCtx, onStopCtxCancel := context.WithCancel(context.Background())
 	// Create the network
 	net := &localNetwork{
 		networkID:          networkID,
 		genesis:            []byte(networkConfig.Genesis),
 		nodes:              map[string]*localNode{},
-		closedOnStopCh:     make(chan struct{}),
+		onStopCtx:          onStopCtx,
+		onStopCtxCancel:    onStopCtxCancel,
 		log:                log,
 		bootstraps:         beacon.NewSet(),
 		newAPIClientF:      newAPIClientF,
@@ -364,7 +376,7 @@ func (ln *localNetwork) AddNode(nodeConfig node.Config) (node.Node, error) {
 
 // Assumes [ln.lock] is held.
 func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
-	if ln.isStopped() {
+	if ln.stopCalled() {
 		return nil, network.ErrStopped
 	}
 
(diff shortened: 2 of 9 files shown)


local/network.go
```diff
@@ -66,8 +66,18 @@ type localNetwork struct {
 	newAPIClientF api.NewAPIClientF
 	// Used to create new node processes
 	nodeProcessCreator NodeProcessCreator
-	// Closed when network is done shutting down
-	closedOnStopCh chan struct{}
+	stopOnce           sync.Once
+	// Cancelled when Stop begins.
+	onStopCtx context.Context
```
felipemadero, line 71: Is using a context more idiomatic than using a channel? I prefer reducing the number of used fields for a given solution. It seems the context is not wrapped by any other context, so a channel can be used instead. Whould like to know you opinion.

local/network.go
```diff
@@ -66,8 +66,18 @@ type localNetwork struct {
...
+	// Cancels [onStopCtx]
+	onStopCtxCancel func()
+	// True if Stop has returned.
+	// This field only exists so that in removeNode,
+	// we know whether Stop has already completed, in
+	// which case we return an error, or not, in which
+	// case we proceed to remove all nodes.
+	// [lock] must be held while accessing this field.
+	stopped bool
```
felipemadero, line 80: This can be avoided is the check for function preconditions (currently mainly stopped network), is moved to the public methods, and so the non public ones can be used at discretion. What do you think?

local/network.go
```diff
@@ -435,59 +447,61 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
...
-			healthyChan <- err
+	if ln.stopCalled() {
+		return network.ErrStopped
+	}
+
+	// Derive a new context that's cancelled when Stop is called,
+	// so that we calls to Healthy() below immediately return.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	go func(ctx context.Context) {
```
felipemadero, line 465: does it need to receive ctx parameter? it is in the scope

local/network.go
```diff
@@ -435,59 +447,61 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
...
+		// This goroutine runs until [ln.Stop] is called
+		// or this function returns.
+		select {
+		case <-ln.onStopCtx.Done():
```
felipemadero, line 469: while not moving this to the next select, reducing the number of goroutines?

local/network.go
```diff
@@ -435,59 +447,61 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
...
+			// Do this until ctx timeout or network closed.
+			for {
+				health, err := node.client.HealthAPI().Health(ctx)
+				if err == nil && health.Healthy {
+					ln.log.Debug("node %q became healthy", node.name)
+					return nil
+				}
+				select {
+				case <-ctx.Done():
+					return fmt.Errorf("node %q failed to become healthy within timeout", node.GetName())
```
felipemadero, line 489: not only failing for timeout, but also for network stop.

local/network.go
```diff
@@ -435,59 +447,61 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
...
+		errGr.Go(func() error {
+			// Every [healthCheckFreq], query node for health status.
+			// Do this until ctx timeout or network closed.
+			for {
+				health, err := node.client.HealthAPI().Health(ctx)
+				if err == nil && health.Healthy {
+					ln.log.Debug("node %q became healthy", node.name)
+					return nil
+				}
+				select {
```
felipemadero, line 487: I shall move `case <-ln.onStopCtx.Done():` to this select. Why not?

local/network.go
```diff
@@ -533,18 +547,22 @@ func (ln *localNetwork) GetAllNodes() (map[string]node.Node, error) {
 }
 
 func (ln *localNetwork) Stop(ctx context.Context) error {
-	ln.lock.Lock()
-	defer ln.lock.Unlock()
+	err := network.ErrStopped
+	ln.stopOnce.Do(
```
felipemadero, line 551: Just a comment on the pattern. I shall prefer something similar to this, but with the error to return after the starting execution included in the `once` call. Somethings like this.
```
return ln.stopOnce(func, afterOnceErr)
```
Of course I understand this does not exits but shall more readable in my opinion.

local/network.go
```diff
@@ -576,11 +594,12 @@ func (ln *localNetwork) RemoveNode(nodeName string) error {
 	return ln.removeNode(nodeName)
 }
 
-// Assumes [net.lock] is held
+// Assumes [ln.lock] is held
 func (ln *localNetwork) removeNode(nodeName string) error {
-	if ln.isStopped() {
+	if ln.stopped {
```
felipemadero, line 599: this can be avoided in the check isStopped() is moved to RemoveNode. Same for all other public functions.

local/network_test.go
```diff
@@ -1216,3 +1216,54 @@ func TestRemoveBeacon(t *testing.T) {
...
+		healthyChan <- net.Healthy(context.Background())
+	}()
+	// Wait to make sure we're actually blocking on Health API call
+	time.Sleep(500 * time.Millisecond)
+	err = net.Stop(context.Background())
+	assert.NoError(err)
+	select {
+	case err := <-healthyChan:
+		assert.Error(err)
+	case <-time.After(3 * time.Second):
```
felipemadero, line 1266: constant for the time? why 3?

server/custom_vms.go
```diff
@@ -91,16 +91,18 @@ func (lc *localNetwork) waitForCustomVMsReady(ctx context.Context) error {
...
-	hc := lc.nw.Healthy(ctx)
-	select {
-	case <-lc.stopc:
-		return errAborted
-	case <-ctx.Done():
-		return ctx.Err()
-	case err := <-hc:
-		if err != nil {
-			return err
+	healthyCtx, healthyCtxCancel := context.WithCancel(ctx)
```
felipemadero, line 94: I think you do not need this. Healthy is already taking into account stop events. You can ignore lc.stopc.

server/network.go
```diff
@@ -284,16 +284,18 @@ var errAborted = errors.New("aborted")
...
-	hc := lc.nw.Healthy(ctx)
-	select {
-	case <-lc.stopc:
-		return errAborted
-	case <-ctx.Done():
-		return ctx.Err()
-	case err := <-hc:
-		if err != nil {
-			return err
+	healthyCtx, healthyCtxCancel := context.WithCancel(ctx)
```
felipemadero, line 287: I think you do not need this. Healthy is already taking into account stop events. You can ignore lc.stopc.

local/network.go
danlaine: Yes, it needs to take `ctx` as an argument. Otherwise there's a race condition because the goroutine references `ctx` but `ctx` is re-assigned on line 475.

local/network.go
danlaine: If we do that, then if `Stop` is called while we're in a call to `node.client.HealthAPI().Health(ctx)`, the context won't be cancelled, and the goroutine won't finish until the call to `Health` times out. The point of this goroutine is to cancel the context passed into `Health` immediately when `network.Stop` is called.

local/network.go
danlaine: See comment above

local/network.go
danlaine: I'm not sure I understand what you mean. Maybe you could paste a snippet with the code you're thinking of?

If you're suggesting that we return from `Stop` inside the anonymous function in `Do`, that isn't possible because returning inside that anonymous function will only return from the anonymous function, not from `Stop`

local/network.go
danlaine: changed to `onStopCh`

local/network_test.go
danlaine: Changed to 1 second (in practice it shouldn't take even that long for us to receive the error on `healthyChan`) and added a comment explaining what this does.

server/network.go
danlaine: removed

local/network.go
danlaine: removed

local/network.go
danlaine: updated comment

local/network.go
danlaine: moved

local/network.go
felipemadero: Ok I see

local/network.go
felipemadero: Ohh something like these 

```
type doOnce struct {                                                                                                        
    mu sync.Mutex                                                                                                           
    done bool                                                                                                               
}                                                                                                                           
                                                                                                                            
func (do doOnce) do(toDo func() error, errAfterOnce error) error {                                                          
    do.mu.Lock()                                                                                                            
    defer do.mu.Unlock()                                                                                                    
    var err error                                                                                                           
    if !do.done {                                                                                                           
        err = toDo()                                                                                                        
        do.done = true                                                                                                      
    } else {                                                                                                                
        err = errAfterOnce                                                                                                  
    }                                                                                                                       
    return err                                                                                                              
}                           
```
Then for field 

`stopOnce doOnce`

Using

 ```
func (ln *localNetwork) Stop(ctx context.Context) error {                                                                   
    return ln.stopOnce.Do(                                                                                                  
        func() error {                                                                                                      
            close(ln.onStopCh)                                                                                              
            ln.lock.Lock()                                                                                                  
            defer ln.lock.Unlock()                                                                                          
            return ln.stop(ctx)                                                                                             
        },                                                                                                                  
        network.ErrStopped,                                                                                                 
    )                                                                                                                       
}              
```

Comment (felipemadero) [reply local/network.go:541]: I understand this really only works for error returns, so probably not general

Comment (danlaine) [reply local/network.go:541]: I think it's cleaner as it is

**Review by felipemadero — COMMENTED** [review]
LGTM

File: README.md [hunk 12]
```diff
@@ -655,16 +655,10 @@ The network runner allows users to interact with an AvalancheGo network using th
 ```go
 // Network is an abstraction of an Avalanche network
 type Network interface {
-  // Returns a chan that is closed when
-  // all the nodes in the network are healthy.
-  // If an error is sent on this channel, at least 1
-  // node didn't become healthy before the timeout.
-  // If an error isn't sent on the channel before it
-  // closes, all the nodes are healthy.
+  // Returns true if all the nodes in the network are healthy.
```
holisticode, line 658: Returns `true`?

local/network.go
```diff
@@ -166,15 +167,15 @@ func (npc *nodeProcessCreator) NewNodeProcess(config node.Config, args ...string
 	if config.RedirectStdout {
 		stdout, err := cmd.StdoutPipe()
 		if err != nil {
-			return nil, fmt.Errorf("Could not create stdout pipe: %s", err)
+			return nil, fmt.Errorf("couldn't create stdout pipe: %s", err)
```
holisticode, line 170: not arguing if `couldn't` is better than `could not` :smile:

local/network.go
```diff
@@ -435,59 +437,61 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
...
+	// Derive a new context that's cancelled when Stop is called,
+	// so that we calls to Healthy() below immediately return.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	go func(ctx context.Context) {
+		// This goroutine runs until [ln.Stop] is called
+		// or this function returns.
+		select {
+		case <-ln.onStopCh:
+			cancel()
```
holisticode, line 460: Thought there could be a problem if `cancel` is called multiple times but it looks there's none

local/network.go
```diff
@@ -603,10 +610,10 @@ func (ln *localNetwork) removeNode(nodeName string) error {
 	return nil
 }
 
-// Assumes [net.lock] is held
-func (ln *localNetwork) isStopped() bool {
+// Returns whether Stop has been called.
+func (ln *localNetwork) stopCalled() bool {
```
holisticode, line 614: I prefer `isStopping` or something like that, reflects more that it is in a certain state

network/network.go
```diff
@@ -11,15 +11,10 @@ var ErrStopped = errors.New("network stopped")
 
 // Network is an abstraction of an Avalanche network
 type Network interface {
-	// Returns a chan that is closed when
-	// all the nodes in the network are healthy.
-	// If an error is sent on this channel, at least 1
-	// node didn't become healthy before the timeout.
-	// If an error isn't sent on the channel before it
-	// closes, all the nodes are healthy.
+	// Returns true if all the nodes in the network are healthy.
```
holisticode, line 14: Returns true?

holisticode (commented):
Looks nice this way.
A couple of nits, specifically  in comments, should be addressed

local/network.go
danlaine: lol I mainly changed this because my linter was complaining about an error message starting with a capital letter

local/network.go
danlaine: Nope, it's fine to do that. Example: https://go.dev/play/p/osLmyXi3y1z

local/network.go
danlaine: I thought about that, but then I didn't name it that because it might actually be _stopped_ as opposed to _stopping_

network/network.go
danlaine: fixed

README.md
danlaine: fixed

holisticode (approved):

PR #158, opened 2022-05-25
Dan's pass; remove unexpectedStopChan

danlaine:
* Move process status into its own package
* Start process inside `NewNodeProcess`
* Remove `Start` from `NodeProcess` interface
* Change guarantees around `NodeProcess`'s  `Wait` and `Stop` methods. 
* Remove process status Initial
* Remove unexpected stop channel


local/network.go
```diff
@@ -560,23 +560,14 @@ func (ln *localNetwork) Stop(ctx context.Context) error {
...
-			// In practice we'll probably never time out here,
-			// and the caller probably won't cancel a call
-			// to stop(), but we include this to respect the
-			// network.Network interface.
-			return ctx.Err()
-		default:
-		}
-		if err := ln.removeNode(ctx, nodeName); err != nil {
+		stopCtx, stopCtxCancel := context.WithTimeout(ctx, stopTimeout)
+		if err := ln.removeNode(stopCtx, nodeName); err != nil {
```
danlaine, line 566: Previously, all calls to `removeNode` shared a context and timeout, so if one node timed out, all other nodes would be SIGKILLed. Now, each node gets its own timeout so this won't happen.

server/server.go
```diff
@@ -887,34 +885,35 @@ func (s *server) getClusterInfo() *rpcpb.ClusterInfo {
 	return info
 }
 
-func (s *server) handleUnexpectedNodeStop(unexpectedNodeStopCh chan network.UnexpectedNodeStopMsg) {
+func (s *server) handleUnhealthy() {
+	timer := time.NewTimer(0)
```
holisticode, line 889: Huh? Why wanting to drain the timer right away if you just created it? I don't think that will ever enter? 
Also, why not just create the timer with the `healthCheckFreq` right away? Creating with 0 doesn't seem to have any effect and it confuses me trying to understand what's going on...

server/server.go
```diff
@@ -887,34 +885,35 @@ func (s *server) getClusterInfo() *rpcpb.ClusterInfo {
...
+			err := s.network.nw.Healthy(ctx)
+			cancel()
+			if err == nil {
+				continue
+			}
 			zap.L().Info(
-				"received unexpected node stop message",
-				zap.String("node-name", unexpectedStopMsg.NodeName),
-				zap.Int("exit-code", unexpectedStopMsg.ExitCode),
+				"network is unhealthy",
```
holisticode, line 912: This actually means that the health call failed, not necessarily that it is unhealthy. It probably is in that case, but I am not sure about the semantics, would this check also have to fail if the network is just unhealthy?

holisticode (commented):
Basically lgtm, there are some small issues I'd like to understand better?

local/network.go
```diff
@@ -467,23 +464,26 @@ func (ln *localNetwork) Healthy(ctx context.Context) error {
...
+		nodeName := node.GetName()
 		errGr.Go(func() error {
 			// Every [healthCheckFreq], query node for health status.
 			// Do this until ctx timeout or network closed.
 			for {
-				if lnNode.Status() != node.Started {
-					return fmt.Errorf("unexpected stop on node %q", lnNode.GetName())
+				if node.Status() != status.Running {
+					// If we had stopped this node ourselves, it wouldn't be in [ln.nodes].
+					// Since it is, it means the node stopped unexpectedly.
```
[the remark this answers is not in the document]
danlaine: Because we hold the lock here, either:
1. `ln.Stop` was previously called, in which case we return `ErrStopped` above.
2. `ln.Stop` wasn't previously called, in which case it can't execute until after `Healthy` returns. So no nodes in `ln.nodes` should be stopped.

local/network_test.go
```diff
@@ -52,12 +53,7 @@ func (*localTestSuccessfulNodeProcessCreator) NewNodeProcess(config node.Config,
 type localTestFailedStartProcessCreator struct{}
 
 func (*localTestFailedStartProcessCreator) NewNodeProcess(config node.Config, flags ...string) (NodeProcess, error) {
-	process := &mocks.NodeProcess{}
-	process.On("Start", mock.Anything).Return(errors.New("Start failed"))
-	process.On("Wait").Return(nil)
-	process.On("Stop", mock.Anything).Return(nil)
-	process.On("Status").Return(node.Started)
-	return process, nil
+	return nil, errors.New("an error")
```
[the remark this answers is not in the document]
danlaine: changed

server/server.go
danlaine: changed the logic here

server/server.go
danlaine: I'm not sure I follow. This is in the server, not the client. `Healthy` returns a non-nil error if a node isn't healthy.

holisticode (approved):

PR #159, opened 2022-05-26
Cancellable start context

holisticode:
Currently the Start context can not be cancelled.
This is required though if for example a client application hits Ctrl-C and wants to clean up immediately

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -52,9 +52,10 @@ type Server interface {
 type server struct {
 	cfg Config
 
-	rootCtx   context.Context
-	closeOnce sync.Once
-	closed    chan struct{}
+	rootCtx    context.Context
+	cancelFunc context.CancelFunc // allow the Start context to be cancelled
+	closeOnce  sync.Once
+	closed     chan struct{}
 
 	ln               net.Listener
 	gRPCServer       *grpc.Server
@@ -182,10 +183,12 @@ func (s *server) Run(rootCtx context.Context) (err error) {
 		s.gRPCServer.Stop()
 		zap.L().Warn("closed gRPC server")
 		<-gRPCErrc
+		zap.L().Warn("gRPC terminated")
 
 	case err = <-gRPCErrc:
 		zap.L().Warn("gRPC server failed", zap.Error(err))
 		if !s.cfg.GwDisabled {
+			zap.L().Warn("closed gRPC gateway server", zap.Error(s.gwServer.Close()))
 			zap.L().Warn("closed gRPC gateway server", zap.Error(s.gwServer.Close()))
 			<-gwErrc
 		}
@@ -200,6 +203,9 @@ func (s *server) Run(rootCtx context.Context) (err error) {
 	if s.network != nil {
 		stopCtx, stopCtxCancel := context.WithTimeout(context.Background(), StopOnSignalTimeout)
 		defer stopCtxCancel()
+		// we need to cancel the Start context so that any pending action with that context
+		// can be cancelled
+		s.cancelFunc()
 		s.network.stop(stopCtx)
 		zap.L().Warn("network stopped")
 	}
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -52,9 +52,10 @@ type Server interface {
 type server struct {
 	cfg Config
 
-	rootCtx   context.Context
-	closeOnce sync.Once
-	closed    chan struct{}
+	rootCtx    context.Context
+	cancelFunc context.CancelFunc // allow the Start context to be cancelled
```
felipemadero, line 56: startCtxCancel ?

server/server.go
```diff
@@ -348,9 +354,14 @@ func (s *server) Start(ctx context.Context, req *rpcpb.StartRequest) (*rpcpb.Sta
 		return nil, err
 	}
 
+	// start triggers a series of different time consuming actions
+	// (in case of subnets: create a wallet, create subnets, issue txs, etc.)
+	// We may need to cancel the context, for example if the client hits Ctrl-C
+	var cancellableCtx context.Context
+	cancellableCtx, s.cancelFunc = context.WithCancel(ctx)
```
felipemadero, line 361: startCtx, startCtxCancel

server/server.go
```diff
@@ -200,6 +203,9 @@ func (s *server) Run(rootCtx context.Context) (err error) {
 	if s.network != nil {
 		stopCtx, stopCtxCancel := context.WithTimeout(context.Background(), StopOnSignalTimeout)
 		defer stopCtxCancel()
+		// we need to cancel the Start context so that any pending action with that context
+		// can be cancelled
+		s.cancelFunc()
```
felipemadero, line 208: we need also to add this cancelation to the Stop() function of the server, so CLI stop command also cancels a probably concurrently running Start(), without needing to shutdown the server

server/server.go
```diff
@@ -182,10 +183,12 @@ func (s *server) Run(rootCtx context.Context) (err error) {
 		s.gRPCServer.Stop()
 		zap.L().Warn("closed gRPC server")
 		<-gRPCErrc
+		zap.L().Warn("gRPC terminated")
 
 	case err = <-gRPCErrc:
 		zap.L().Warn("gRPC server failed", zap.Error(err))
 		if !s.cfg.GwDisabled {
+			zap.L().Warn("closed gRPC gateway server", zap.Error(s.gwServer.Close()))
```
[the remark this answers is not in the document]
holisticode: no idea, must have been some IDE issue, fixing

server/server.go
```diff
@@ -200,6 +202,9 @@ func (s *server) Run(rootCtx context.Context) (err error) {
 	if s.network != nil {
 		stopCtx, stopCtxCancel := context.WithTimeout(context.Background(), StopOnSignalTimeout)
 		defer stopCtxCancel()
+		// we need to cancel the Start context so that any pending action with that context
+		// can be cancelled
+		s.startCtxCancel()
```
felipemadero, line 207: as there are two ways to call network.stop() here (server receiving SIGINT or server.Stop() call), maybe this start context creation can be moved down to network.start() and the start context cancelation can be moved down to network.stop() so network struct manages that logic

server/server.go
```diff
@@ -772,6 +782,10 @@ func (s *server) Stop(ctx context.Context, req *rpcpb.StopRequest) (*rpcpb.StopR
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	// cancel possible concurrent still running start
```
felipemadero, line 785: this is somewhat related, probably check for started cluster `if info == nil` shall be removed from Stop(), so Stop() can be called even if the first health check had not returned yet

holisticode: I am not sure I fully understood what you mean, and also, hope there is no other consequence, but I provided an update to this thinking to have addressed it.

server/server.go
holisticode: If I understood you correctly here this is a good suggestion. Changed.

felipemadero (approved):

felipemadero (approved):

PR #167, opened 2022-06-14
Replace random log names with timestamps

dasconnor:
Log and root data dirs will now be named with timestamps instead of with random strings. Root data dir can now also be set when loading a snapshot.


client/client.go
```diff
@@ -43,7 +43,7 @@ type Client interface {
 	SendOutboundMessage(ctx context.Context, nodeName string, peerID string, op uint32, msgBody []byte) (*rpcpb.SendOutboundMessageResponse, error)
 	Close() error
 	SaveSnapshot(ctx context.Context, snapshotName string) (*rpcpb.SaveSnapshotResponse, error)
-	LoadSnapshot(ctx context.Context, snapshotName string) (*rpcpb.LoadSnapshotResponse, error)
+	LoadSnapshot(ctx context.Context, snapshotName string, rootDataDir string) (*rpcpb.LoadSnapshotResponse, error)
```
felipemadero, line 46: I shall prefer to pass rootDataDir as optional argument, as Start() does (same as execPath and pluginDir)

felipemadero: See for that the updated code for LoadSnapshot, at current main

cmd/avalanche-network-runner/control/control.go
```diff
@@ -636,10 +636,10 @@ func saveSnapshotFunc(cmd *cobra.Command, args []string) error {
 
 func newLoadSnapshotCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "load-snapshot snapshot-name",
+		Use:   "load-snapshot snapshot-name [root-data-dir]",
 		Short: "Requests server to load network snapshot.",
 		RunE:  loadSnapshotFunc,
-		Args:  cobra.ExactArgs(1),
+		Args:  cobra.RangeArgs(1, 2),
```
felipemadero, line 642: root data dir should be passed optional, with a flag, same as start, and other parameter for load snapshot in current main

client/client.go
dasconnor: fixed

cmd/avalanche-network-runner/control/control.go
dasconnor: done

server/server.go
```diff
@@ -1023,7 +1029,12 @@ func (s *server) LoadSnapshot(ctx context.Context, req *rpcpb.LoadSnapshotReques
 		err error
 	)
 
-	rootDataDir, err := os.MkdirTemp(os.TempDir(), "network-runner-root-data")
+	rootDataDir := req.GetRootDataDir()
+	if len(rootDataDir) == 0 {
+		rootDataDir = "/tmp"
```
felipemadero, line 1034: Probably shall be os.TempDir()

dasconnor: fixed

felipemadero (approved):

PR #171, opened 2022-06-27
change format for root dir

felipemadero:
- change timestamp so shells expansion work best (without : in time)
- unify root dir format for both local and server

### DIFF
--- a/local/network.go
+++ b/local/network.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"path"
 	"path/filepath"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


local/network.go
```diff
@@ -241,7 +242,8 @@ func newNetwork(
 ) (*localNetwork, error) {
 	var err error
 	if rootDir == "" {
-		rootDir, err = os.MkdirTemp("", fmt.Sprintf("%s-*", rootDirPrefix))
+		rootDir = path.Join(os.TempDir(), rootDirPrefix)
```
danlaine, line 245: We should use [filepath.Join](https://pkg.go.dev/path/filepath) instead of `path.Join`. 

From https://pkg.go.dev/path:

```
The path package should only be used for paths separated by forward slashes, such as the paths in URLs. This package does not deal with Windows paths with drive letters or backslashes; to manipulate operating system paths, use the path/filepath package.
```

dasconnor (approved):
LGTM

PR #174, opened 2022-07-04
AV-1868 add call to LoadVMs for all nodes, so as to load new binaries if present

felipemadero:
- dynamically load VMs binaries by using admin API


dasconnor (approved):
LGTM

PR #182, opened 2022-07-18
Added version and moved to CLI format

### LINKED ISSUE #181: Add version command to ANR
Add version command to avalanche-network-runner so that the version can be printed from the tool in the same way that it’s done with the avalanche-cli.

dasconnor:
I added a version command and made the repo a bit more like avalanche-cli.

To do that, I
- Created a `main.go` to sit at the repo's top level. This calls the root command.
- The root command is now in the `cmd` package, not `main`
- I removed an unnecessary folder in the `cmd/` directory
- Added a build script

Closes #181


VERSION
```diff
@@ -0,0 +1 @@
+0.1.4
```
holisticode, line 1: Isn't this `1.1.4`?

scripts/tests.e2e.sh
```diff
@@ -102,7 +102,7 @@ rm -rf $snapshots_dir
 killall network.runner || echo
 
 echo "launch local test cluster in the background"
-/tmp/network.runner \
+bin/avalanche-network-runner \
```
holisticode, line 105: A minor thing, where there's no real consensus I guess: I try to avoid to generate stuff into the working directory. For example, when moving around (copy) the whole directory (for example when copying to an experimental instance, where we can't use git), it gets copied with it, even if it's in `.gitignore`...

VERSION
dasconnor: fixed

scripts/tests.e2e.sh
dasconnor: I guess I'm of the opposing philosophy. Generally, if a script generates something, it should be in the same working directory. If you wanted to copy the directory cleanly, you could just clone again. If we did want to change this, I think the better idea would be to adjust the build script to accept a custom install location as a command line flag. But default should still be bin.

dasconnor: Ok, I went ahead and changed the build script. Now you can build to tmp if you wish, but it won't be default

holisticode: That's very nice, thanks.

holisticode (approved):

Issue #187, opened 2022-07-19
Move ANR AttachPeer + SendOutboundmessage extra logic from server to local (was [AV-1929])


holisticode:
Was originally created at https://ava-labs.atlassian.net/browse/AV-1929

This functions should be as short as possible

PR #188, opened 2022-07-19
Move AttachPeer entirely to the node

holisticode:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/187 

Moved all AttachPeer stuff from the network to the local node.


local/node.go
```diff
@@ -159,13 +161,26 @@ func (node *localNode) AttachPeer(ctx context.Context, router router.InboundHand
 			peerMsgQueueBufferSize,
 		),
 	)
+	cctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
felipemadero, line 164: constant for this?

felipemadero (commented):
LGTM besides my comments and connor's one

server/server.go
```diff
@@ -932,9 +944,9 @@ func (s *server) AttachPeer(ctx context.Context, req *rpcpb.AttachPeerRequest) (
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	node, err := s.network.nw.GetNode((req.NodeName))
+	node, err := s.network.nw.GetNode(req.NodeName)
 	if err != nil {
-		return nil, err
+		return nil, ErrNodeNotFound
```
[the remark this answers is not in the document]
holisticode: I thought as we are getting a node, and it failed, we could just return ErrNodeNotFound. But yes, it's probably better to return the actual error. Fixed.

local/node.go
holisticode: Added

felipemadero (approved):

PR #196, opened 2022-07-21
Move restart to local

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/192


local/network.go
```diff
@@ -706,6 +706,49 @@ func (ln *localNetwork) removeNode(ctx context.Context, nodeName string) error {
...
+
+	if dbDir != "" {
+		nodeConfig.Flags[config.DBPathKey] = dbDir
+	}
+
+	if err := ln.removeNode(ctx, nodeName); err != nil {
+		return err
+	}
+
+	if _, err := ln.addNode(nodeConfig); err != nil {
```
felipemadero, line 745: is this preserving ports?

local/network.go
```diff
@@ -714,6 +714,55 @@ func (ln *localNetwork) removeNode(ctx context.Context, nodeName string) error {
...
+	if binaryPath != "" {
+		nodeConfig.BinaryPath = binaryPath
+		nodeConfig.Flags[config.BuildDirKey] = filepath.Dir(binaryPath)
+	}
+
+	if whitelistedSubnets != "" {
+		nodeConfig.Flags[config.WhitelistedSubnetsKey] = whitelistedSubnets
+	}
+
+	if dbDir != "" {
```
holisticode, line 745: Why would restarting a node actually allow to change the dbDir? Isn't it the same node?

local/network.go
felipemadero: it is

local/network.go
felipemadero: double checked. currently there is no need to accept dbdir, rootdir as param to restart.

holisticode (approved):

PR #200, opened 2022-07-22
189 addnode to local e2e fix

felipemadero:


server/network.go
```diff
@@ -156,19 +154,34 @@ func (lc *localNetwork) createConfig() error {
...
-		mergedConfig, err := mergeNodeConfig(local.GetDefaultFlags(), globalConfig, lc.options.customNodeConfigs[nodeName])
-		if err != nil {
-			return fmt.Errorf("failed merging provided configs: %w", err)
+		if cfg.NodeConfigs[i].Flags == nil {
+			cfg.NodeConfigs[i].Flags = map[string]interface{}{}
+		}
+
+		// set flags applied to the specific node
+		var customNodeConfig map[string]interface{}
+		if lc.options.customNodeConfigs[nodeName] != "" {
```
holisticode, line 163: Are we sure that lc.options.customNodeConfigs can never be nil here (to avoid a potential panic)?

server/network.go
```diff
@@ -156,19 +154,34 @@ func (lc *localNetwork) createConfig() error {
...
-		cfg.NodeConfigs[i].ConfigFile, err = createConfigFileString(mergedConfig, logDir, dbDir, buildDir, lc.options.whitelistedSubnets)
-		if err != nil {
-			return err
+
+		cfg.NodeConfigs[i].Flags[config.LogsDirKey] = logDir
+		cfg.NodeConfigs[i].Flags[config.DBPathKey] = dbDir
+		if buildDir != "" {
+			cfg.NodeConfigs[i].Flags[config.BuildDirKey] = buildDir
+		}
+		if lc.options.whitelistedSubnets != "" {
```
holisticode, line 183: Same here with `whitelistedSubnets`

server/server.go
```diff
@@ -824,22 +823,18 @@ func (s *server) RestartNode(ctx context.Context, req *rpcpb.RestartNodeRequest)
...
-	)
-	if err != nil {
-		return nil, fmt.Errorf("failed to generate json node config string: %w", err)
+	nodeConfig.Flags[config.LogsDirKey] = nodeInfo.LogDir
+	nodeConfig.Flags[config.DBPathKey] = nodeInfo.DbDir
+	if buildDir != "" {
+		nodeConfig.Flags[config.BuildDirKey] = buildDir
+	}
+	if nodeInfo.WhitelistedSubnets != "" {
+		nodeConfig.Flags[config.WhitelistedSubnetsKey] = nodeInfo.WhitelistedSubnets
```
holisticode, line 837: Just wondering: should the restarting node keep apiPort and p2pPort?

holisticode (commented):
Just questions and nits, otherwise looks good to me

server/network.go
felipemadero: not sure. adding guard

server/network.go
felipemadero: the string can't be nil

server/server.go
felipemadero: Yes, it should. If you think it is not doing, pls add this comment also to the restart node migration PR so we can fix the problem there

server/network.go
holisticode: I meant `lc.options` could?

felipemadero: per the current sequence of calls, not

server/server.go
holisticode: Seems to be addressed here: https://github.com/ava-labs/avalanche-network-runner/pull/196#discussion_r932603987

holisticode (approved):

PR #201, opened 2022-07-27
add a pair of public errors and a function to check error kind

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/202


server/server.go
```diff
@@ -93,6 +94,11 @@ const (
 	rootDataDirPrefix = "network-runner-root-data"
 )
 
+func IsServerError(err error, serverError error) bool {
```
holisticode, line 97: Why add this function, where is this used?

server/server.go
```diff
@@ -93,6 +94,11 @@ const (
...
+	status := status.Convert(err)
+	return status.Code() == codes.Unknown && status.Message() == serverError.Error()
```
holisticode, line 99: It's not immediately obvious why these comparisons result in a server error, could you add some documentation?

server/server.go
felipemadero: It is used by the CLI, or other server clients, to check for error type of server return. It is the current proposal to solve the issue of the rpc implementation destroying the error type (by encapsulating msg string inside of status type). Other proposal should be for the client.Client to get error msgs inside status, and returning the appropiate server type.

server/server.go
felipemadero: added:
```
// grpc encapsulates the non protocol-related, ANR server domain errors,
// inside grpc.status.Status structs, with status.Code() code.Unknown,
// and original error msg inside status.Message() string
// this aux function is to be used by clients, to check for the appropiate
// ANR domain error kind
```

holisticode (approved):

Issue #205, opened 2022-07-28
fix obsolete/unclear ANR server CustomVMS issues [AV-1918]


felipemadero:
- avoid using the same name custom_vms both for user input and user output at RPC (different types and contents)
- fix comment on ClusterInfo.CustomVMS in rpc proto. currently the index is blockchain ID
- rename ClusterInfo.CustomVMS.VmName to ClusterInfo.CustomVMS.BlockchainName. it is technically that
- add comments to make clear that Vm Name is beign saved as blockchain name

PR #206, opened 2022-07-28
Fix unclear rpc issues

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/205


rpcpb/rpc.proto
```diff
@@ -152,15 +152,18 @@ message ClusterInfo {
   // Maps from the node ID to its attached peer infos.
   map<string, ListOfAttachedPeerInfo> attached_peer_infos = 6;
 
-  // Set to "true" once custom VMs are ready.
-  bool custom_vms_healthy = 7;
-  // The map of custom VM IDs in "ids.ID" format to its VM information.
-  map<string, CustomVmInfo> custom_vms = 8;
+  // Set to "true" once custom blockchains are ready.
+  bool custom_chains_healthy = 7;
+  // The map of blockchain IDs in "ids.ID" format to its information.
```
holisticode, line 157: Nit: this sentence sounds strange now, "to its information" refers to what?

holisticode (commented):
Basically lgtm but I have two main questions:
1. The renaming from VM to chains in the PR seems consistent, but has this been discussed and approved? I mean, I remember a lot of mentions of VMs in our docs, not necessarily chains. Are we sure we want to always use chain now?

2. Does this not break a lot of things dependent on ANR? For example, I think we check `CustomVmInfo` inside avalanche-cli for example

rpcpb/rpc.proto
felipemadero: went back to previous sentence, replacing vm with blockchain

felipemadero:
> Basically lgtm but I have two main questions:
> 
> 1. The renaming from VM to chains in the PR seems consistent, but has this been discussed and approved? I mean, I remember a lot of mentions of VMs in our docs, not necessarily chains. Are we sure we want to always use chain now?
> 2. Does this not break a lot of things dependent on ANR? For example, I think we check `CustomVmInfo` inside avalanche-cli for example

Probably we want ANR to be as technical as avalanchego regarding the meaning of subnet, vm, blockchain. The changes go in that direction. What do you think of this @dasconnor  ?

Yes, this changes, together with other cleanup RPC changes, break things dependent on ANR. 
At least CLI and subnet-evm will need fixes.

dasconnor:
I think I'm in favor of Felipe's suggestion, we should be as precise with our language as possible. I'm ok with breaking changes, we'll just need to let the subnet-evm team know. It should be a pretty small refactor, right?

dasconnor:
But I guess we should be a little careful, I'm not sure I agree with all the usages of `chain` vs `vm`, let me do a full review

local/blockchain.go
```diff
@@ -50,10 +50,10 @@ var (
 )
 
 type blockchainInfo struct {
-	vmName       string
-	vmID         ids.ID
-	subnetID     ids.ID
-	blockchainID ids.ID
```
dasconnor, line 56: This value is a bit ambiguous to me. To me, `chainID` is the evm parameter (like 43114 for c-chain) and `blockchainID` is the p-chain parameter. If we want the p-chain parameter, we should keep this as `blockchainID`

README.md
```diff
@@ -460,7 +460,7 @@ avalanche-network-runner server \
 curl -X POST -k http://localhost:8081/v1/ping -d ''
 ```
 
-To start the cluster with custom VMs:
+To start the cluster with custom chains:
```
dasconnor, line 463: I don't think `custom VMs` means the same thing as `custom chains`. I think `custom VM` means a non-subnet-evm or subnet-evm fork VM. I'm not sure if we want this to mean all vm types or just the non-subnet-evm ones

local/blockchain.go
```diff
@@ -215,10 +215,12 @@ func (ln *localNetwork) installCustomVMs(
...
-			vmName:       chainSpec.VmName,
-			vmID:         vmID,
-			subnetID:     subnetID,
-			blockchainID: blockchainIDs[i],
+			// we keep a record of VM name in blockchain name field,
+			// as there is no way to recover VM name from VM ID
+			chainName: chainSpec.VmName,
+			vmID:      vmID,
+			subnetID:  subnetID,
+			chainID:   blockchainIDs[i],
```
dasconnor, line 223: same comment here

README.md
dasconnor: We might want to just call this `chain` and not `custom chain`.

felipemadero:
> I think I'm in favor of Felipe's suggestion, we should be as precise with our language as possible. I'm ok with breaking changes, we'll just need to let the subnet-evm team know. It should be a pretty small refactor, right?

Yes. I is going to be mainly renaming. We may also made the PR for subnet-evm changes.

README.md
felipemadero: ok, in order to be uniform, maybe I need to make such changes (just chain, not custom chain) on:
- some logs, vars, function names in local/blockchain.go, server/network.go, server/server.go
-  rpc related: custom_chains_healthy, custom_chains, CustomChainInfo

Do you agree? Do you also agree @holisticode  ?

holisticode: Well to me it sounds odd to just say "To start the cluster with chains". It's nuanced, because, yes, we could say a `custom VM` is a VM not provided by us, so a subnet-evm would not be a custom VM. However, a chain running with the subnet-evm is still a custom chain, as it is not a chain from our primary chains. In fact, if the user provides custom precompiles, a subnet-evm ceases to be vanilla and becomes a customized chain indeed...

felipemadero: will keep as it is, based on private communications

holisticode (approved):

PR #208, opened 2022-07-28
Use buildir on rpc

felipemadero:


holisticode (commented):
lgtm, but this might also result in breaking changes, so we should check on dependent repos (e.g. avalanche-cli)

holisticode (approved):

PR #210, opened 2022-07-29
Attempt to add upgradeConfig support

darioush:
This will help with testing upgradeBytes and possibly using this in `avalanche-cli`


cmd/control/control.go
```diff
@@ -507,6 +508,12 @@ func newAddNodeCommand() *cobra.Command {
 		"",
 		"[optional] JSON string of map that maps from chain id to its config file contents",
 	)
+	cmd.PersistentFlags().StringVar(
+		&upgradeConfigs,
+		"upgrade-configs",
+		"",
+		"[optional] JSON string of map that maps from chain id to its upgrade file contents",
```
holisticode, line 515: Nit: "JSON string of map from chain id to..."

cmd/control/control.go
```diff
@@ -587,6 +602,18 @@ func newRestartNodeCommand() *cobra.Command {
 		"",
 		"whitelisted subnets (comma-separated)",
 	)
+	cmd.PersistentFlags().StringVar(
+		&chainConfigs,
+		"chain-configs",
+		"",
+		"[optional] JSON string of map that maps from chain id to its config file contents",
```
holisticode, line 609: same

cmd/control/control.go
```diff
@@ -587,6 +602,18 @@ func newRestartNodeCommand() *cobra.Command {
...
+	)
+	cmd.PersistentFlags().StringVar(
+		&upgradeConfigs,
+		"upgrade-configs",
+		"",
+		"[optional] JSON string of map that maps from chain id to its upgrade file contents",
```
holisticode, line 615: ditto

cmd/control/control.go
holisticode: I thought chain configs were already supported, @felipemadero can you confirm?

local/network.go
```diff
@@ -1099,17 +1109,23 @@ func (ln *localNetwork) buildFlags(
 		return buildFlagsReturn{}, err
 	}
 
-	// Use random free API port unless given in config file
-	apiPort, err := getPort(nodeConfig.Flags, configFile, config.HTTPPortKey)
-	if err != nil {
-		return buildFlagsReturn{}, err
+	// Prefer API port and P2P (staking) port from nodeConfig
+	apiPort, p2pPort := nodeConfig.ApiPort, nodeConfig.StakingPort
```
holisticode, line 1113: What if those ports were already taken?

holisticode (commented):
To me it looks basically good, I have only a couple of nits, thanks for this addition.

I'd like @felipemadero to check on the chain config changes, as I thought they were already supported.

felipemadero (commented):
Could you update the PR to latest changes? Will do a review tomorrow

local/network.go
darioush: The ports would be used by the same node previously, so it seems reasonable to assume they are available after the node shuts down?
Not sure if you already handled this is a different way but it is useful eg to not have to reconfigure metamask or other clients to submit tx after restarting the node.

felipemadero: Yeah it is already handled in other way in ANR

cmd/control/control.go
```diff
@@ -163,6 +164,12 @@ func newStartCommand() *cobra.Command {
 		"",
 		"[optional] JSON string of map from chain id to its config file contents",
 	)
+	cmd.PersistentFlags().StringVar(
+		&upgradeConfigs,
+		"upgrade-configs",
+		"",
+		"[optional] JSON string of map from chain id to its config file contents",
```
felipemadero, line 171: nit: upgrade file contents

PR #212, opened 2022-08-01
Anr rpc cleanup

felipemadero:


cmd/control/control.go
```diff
@@ -116,10 +116,10 @@ func newStartCommand() *cobra.Command {
 		"[optional] root data directory to store logs and configurations",
 	)
 	cmd.PersistentFlags().StringVar(
-		&customVMNameToGenesisPath,
-		"custom-vms",
+		&blockchainSpecsStr,
+		"blockchain-specs",
 		"",
-		"[optional] JSON string of map that maps from VM to its genesis file path",
+		"[optional] JSON string of list of [(VM name, its genesis file path)]",
```
holisticode, line 122: Nit: Technically in JSON it's an array

rpcpb/rpc.proto
```diff
@@ -206,7 +206,7 @@ message StartRequest {
   // Must be non-empty to install custom VMs.
   optional string plugin_dir = 6;
 
-  // The map of custom VM name to its genesis file path.
+  // The list of custom VM name, its genesis file path, and (optional) subnet id to use.
```
holisticode, line 209: Here too, it's a list of BlockhainSpec.... If you don't want to change all these it's fine

server/server.go
```diff
@@ -248,9 +248,14 @@ func (s *server) Start(ctx context.Context, req *rpcpb.StartRequest) (*rpcpb.Sta
 		pluginDir = filepath.Join(filepath.Dir(req.GetExecPath()), "plugins")
 	}
 	chainSpecs := []blockchainSpec{}
-	if len(req.GetCustomVms()) > 0 {
+	if len(req.GetBlockchainSpecs()) > 0 {
 		zap.L().Info("plugin dir", zap.String("plugin-dir", pluginDir))
-		for vmName, vmGenesisFilePath := range req.GetCustomVms() {
+		for i := range req.GetBlockchainSpecs() {
+			vmName := req.GetBlockchainSpecs()[i].VmName
```
holisticode, line 254: Would you consider `spec := req.GetBlockchainSpecs()[i]` and then just use that?

server/server.go
```diff
@@ -248,9 +248,14 @@ func (s *server) Start(ctx context.Context, req *rpcpb.StartRequest) (*rpcpb.Sta
...
+			vmGenesisFilePath := req.GetBlockchainSpecs()[i].Genesis
+			if req.GetBlockchainSpecs()[i].SubnetId != nil {
```
holisticode, line 256: Maybe this could be checked first if it's mandatory

holisticode (commented):
I have only a couple of nits, otherwise lgtm!

rpcpb/rpc.proto
felipemadero: I prefer not for this

server/server.go
felipemadero: moved to the start of loop

holisticode (approved):

PR #213, opened 2022-08-01
Remove the refreshableWallet and use avalanchego's wallet

holisticode:


local/blockchain.go
```diff
@@ -137,7 +139,7 @@ func (ln *localNetwork) installCustomVMs(
 			if _, err := platformvm.Codec.Unmarshal(subnetTxBytes, &subnetTx); err != nil {
 				return nil, fmt.Errorf("couldn not unmarshal tx for subnet %q: %w", subnetID.String(), err)
 			}
-			pTXs[subnetID] = &subnetTx
+			pTXs = append(pTXs, subnetID)
```
felipemadero, line 142: it seems most part of this for is not needed anymore (obtaining the tx, unmarshall, etc)

felipemadero: probably update the comments and also the var name with txIDs ? subnetTxIDs ?

local/blockchain.go
```diff
@@ -295,33 +298,39 @@ func (ln *localNetwork) setupWalletAndInstallSubnets(
...
 		println()
 		color.Outf("{{green}}reconnecting the wallet client after restart{{/}}\n")
 		clientURI, err := ln.getClientURI()
 		if err != nil {
-			return nil, err
+			return nil, nil, err
+		}
+		testKeychain := secp256k1fx.NewKeychain(genesis.EWOQKey)
+		allTxs := append(pTXs, subnetIDs...)
+		baseWallet, err = primary.NewWalletWithTxs(ctx, clientURI, testKeychain, allTxs...)
```
felipemadero, line 324: is it possible to reuse setupWallet here?

local/blockchain.go
```diff
@@ -249,7 +252,7 @@ func (ln *localNetwork) setupWalletAndInstallSubnets(
 	}
 	platformCli := platformvm.NewClient(clientURI)
 
-	pTXs := make(map[ids.ID]*txs.Tx)
+	pTXs := make([]ids.ID, 0)
```
felipemadero, line 255: []ids.ID{} ?

local/blockchain.go
holisticode: You're very right with the first comment, thanks.
Regarding the second, I'd like to leave it as subnetID, it's a good reminder that the subnet ID is actually also the tx ID. Finally, pTXs was also used inside avalanchego, so I'd like to keep it this way.

local/blockchain.go
holisticode: ok

local/blockchain.go
holisticode: It is, but setupWallet is also doing more stuff like checking balance which seems unnecessary here

Issue #216, opened 2022-08-02
Cleanup of Control command (was [AV-1768])


holisticode:
Original issue: https://ava-labs.atlassian.net/browse/AV-1768

Review all control commands to use args where possible, and mark required where needed

PR #221, opened 2022-08-02
Mark required flags and set node-name as arg not flag

holisticode:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/216


cmd/control/control.go
```diff
@@ -640,17 +618,12 @@ var (
 
 func newSendOutboundMessageCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "send-outbound-message [options]",
-		Short: "Sends an outbound message to an attached peer.",
-		RunE:  sendOutboundMessageFunc,
-		Args:  cobra.ExactArgs(0),
+		Use:       "send-outbound-message [options]",
```
dasconnor, line 621: I think you're missing a `node-name` here

dasconnor (commented):
I think there's one doc string error, then lgtm

cmd/control/control.go
holisticode: you're right, fixed.

felipemadero (approved):

PR #227, opened 2022-08-04
replace build-dir with plugin-dir

felipemadero:


server/network.go
```diff
@@ -94,12 +99,20 @@ func newLocalNetwork(opts localNetworkOptions) (*localNetwork, error) {
 		return nil, err
 	}
 
+	pluginDir := ""
+	if opts.execPath != "" {
```
dasconnor, line 103: Does it matter which of these has precedence?

local/network.go
```diff
@@ -452,6 +454,9 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
 	}
 	addNetworkFlags(ln.log, ln.flags, nodeConfig.Flags)
 
+	// TODO: remove this when plugin-dir is available
+	delete(nodeConfig.Flags, PluginDirKey)
```
dasconnor, line 458: What's the purpose of this line?

dasconnor (commented):
We also need to adjust the ci config to remove the `tag` starter

local/network.go
felipemadero: to avoid giving the plugin-dir param to avalanchego. current avalanchego does not accepts the flag, and makes CI to fail

server/network.go
felipemadero: yes. opts.pluginDir has precedence over opts.execPath. it enables to give some exec path and some other non standard plugin dir.

local/network.go
felipemadero: I understand with these, plugin-dir can not be used in ANR. But I think this does not matter and can make that avalanchego branch to pass CI (what was asked was not to use build-dir which is completely avoided, while doing that, also this PR partially prepares for using plugin-dir flag).

felipemadero: Is this correct @danlaine ?

danlaine: This branch should assume that `build-dir` is no longer a flag, and that the user may pass a flag `plugin-dir`.  So I'm not sure that we want to delete this flag here

felipemadero: Ok I did not want to change certain things because by default ANR creates plugin-dir flags without being requested by the user, so making ANR CI to fail. Ok will revert this is the flag in needed to be used on ANR PR, and comment other parts.

dasconnor (approved):
LGTM, but Fabio or Dan should probably also have a look

PR #232, opened 2022-08-05
bump go version

felipemadero:


.github/workflows/build-test-release.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
-          go-version: "1.17.2" # The Go version to use.
+          go-version: 1.18
```
holisticode, line 41: Wondering if we should put 1.18.4 here

dasconnor (approved):
LGTM

.github/workflows/build-test-release.yml
felipemadero: It seems consistent to use the same go-version spec in all actions

PR #230, opened 2022-08-04
Unify ANR logging

holisticode:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/225

PR #235, opened 2022-08-10
changes for testing statesync on subnet-evm

darioush:
Encountered the following using ANR to test statesync on subnet-evm:
- Allow eth client to connect to different chain IDs
- Allow restart command to specify chain configs
- bump avalanchego version and fix logging types.


cmd/control/control.go
```diff
@@ -552,6 +552,12 @@ func newRestartNodeCommand() *cobra.Command {
 		"",
 		"whitelisted subnets (comma-separated)",
 	)
+	cmd.PersistentFlags().StringVar(
```
holisticode, line 555: Wasn't this already done?

examples/local/fivenodenetwork/main.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/ava-labs/avalanche-network-runner/local"
 	"github.com/ava-labs/avalanche-network-runner/network"
 	"github.com/ava-labs/avalanchego/utils/logging"
+	"go.uber.org/zap"
```
holisticode, line 15: Note: in this PR https://github.com/ava-labs/avalanche-network-runner/pull/230 we are unifying ANR logging to use just the avalanchego logger (which also used zap, but behind), so depending on what gets merged first, someone might need to address.

examples/local/indepth/main.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/ava-labs/avalanchego/config"
 	"github.com/ava-labs/avalanchego/staking"
 	"github.com/ava-labs/avalanchego/utils/logging"
+	"go.uber.org/zap"
```
holisticode, line 17: Same here

local/helpers.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/ava-labs/avalanche-network-runner/network/node"
 	"github.com/ava-labs/avalanchego/config"
 	"github.com/ava-labs/avalanchego/utils/logging"
+	"go.uber.org/zap"
```
holisticode, line 11: And here

rpcpb/rpc.proto
```diff
@@ -300,6 +300,8 @@ message RestartNodeRequest {
   // Optional fields are set to the previous values if empty.
   optional string exec_path           = 2;
   optional string whitelisted_subnets = 3;
+
+  map<string, string> chain_configs = 4;
```
holisticode, line 304: I am a bit confused, hadn't this already been done @felipemadero

holisticode (commented):
basically lgtm!

But:
* PR https://github.com/ava-labs/avalanche-network-runner/pull/230 is unifying ANR logging to just use the avalanchego logger, and this PR is introducing the zap logger in many places where it was not. Most probably these changes will have to be reverted.
* I am a bit unclear about the changes with chainConfigs, I thought they had been done already (remember to have reviewed something already). Maybe @felipemadero knows better.

rpcpb/rpc.proto
darioush: this is supported on node start but not sure how it works if you want to restart a node with a different config.

felipemadero: Ohh the param is absent from restart node, yes, this should be here. Will also do a review of all of this by tomorrow.

cmd/control/control.go
felipemadero: it was not

examples/local/fivenodenetwork/main.go
felipemadero: probably we may accept this, 230, and later make a new PR for introducing zap types

felipemadero (approved):
LGTM

PR #236, opened 2022-08-11
update cchain eth-api flag names

darioush:
The previous names have been replaced in 0.8.5 with the new alternatives.
Deprecated names are still supported but we should change them here in case we drop support for them later.


felipemadero (approved):
LGTM

holisticode (approved):

PR #244, opened 2022-08-25
add dependabot and codeql to CI

felipemadero:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
```
holisticode, line 34: Sidenote: a `matrix` is not strictly necessary if there is only one element.

holisticode (approved):

Issue #311, opened 2022-10-20
directly use a marshalled string of AvalancheLocalChainConfig 


felipemadero:
keep updated the genesis by using AvalancheLocalChainConfig from coreth and inserting it in each used genesis

PR #315, opened 2022-10-21
Use cchain config from coreth instead of hardcoded

holisticode:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/311


network/config.go
```diff
@@ -15,19 +15,27 @@ import (
...
+	defaultCChainConfigStr = `{"config":{"chainId":43112,"homesteadBlock":0,"daoForkBlock":0,"daoForkSupport":true,"eip150Block":0,"eip150Hash":"0x2086799aeebeae135c246c65021c82b4e15a2c451340993aacfd2751886514f0","eip155Block":0,"eip158Block":0,"byzantiumBlock":0,"constantinopleBlock":0,"petersburgBlock":0,"istanbulBlock":0,"muirGlacierBlock":0,"apricotPhase1BlockTimestamp":0,"apricotPhase2BlockTimestamp":0,"apricotPhase3BlockTimestamp":0,"apricotPhase4BlockTimestamp":0,"apricotPhase5BlockTimestamp":0},"nonce":"0x0","timestamp":"0x0","extraData":"0x00","gasLimit":"0x5f5e100","difficulty":"0x0","mixHash":"0x0000000000000000000000000000000000000000000000000000000000000000","coinbase":"0x0000000000000000000000000000000000000000","number":"0x0","gasUsed":"0x0","parentHash":"0x0000000000000000000000000000000000000000000000000000000000000000"}`
```
felipemadero, line 25: maybe you can also take this from `local/default/genesis.json` ? probably we can have a function that returns a cChainConf and use it whenever needed?

local/network.go
```diff
@@ -158,6 +159,28 @@ func init() {
 			}
 		}
 	}
+
+	// set the cchain genesis directly from coreth
+	// the whole of `cChainGenesis` should be set as a string, not a json object...
+	corethCChainGenesis := coreth_params.AvalancheLocalChainConfig
+	// but the part in coreth is only the "config" part.
+	// In order to set it easily, first we get the cChainGenesis item
+	cChainGenesis := genesisMap["cChainGenesis"]
```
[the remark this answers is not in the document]
holisticode: I'd rather not. Those genesis are not similar today either, and it requires to maintain a rather awkwardly formatted json packed in a string which requires loads of quoting. 

Furthermore, we are substituting parts of that json in this code, and being able to directly access its map is straight forward.

network/config.go
holisticode: I can do this, I thought that for some reason these needed to be different? But if not, I can look at this being handled the same.

utils/chains.go
```diff
@@ -0,0 +1,44 @@
...
+	// set the cchain genesis directly from coreth
+	// the whole of `cChainGenesis` should be set as a string, not a json object...
+	corethCChainGenesis := coreth_params.AvalancheLocalChainConfig
+	// but the part in coreth is only the "config" part.
+	// In order to set it easily, first we get the cChainGenesis item
+	// convert it to a map
+	cChainGenesisMap, ok := cChainGenesis.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf(
+			"expected field 'cChainGenesis' of genesisMap to be a map[string]interface{}, but it failed with type %T", cChainGenesisMap)
```
felipemadero, line 32: should be printing the type of cChainGenesis ?

utils/chains.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+
+	"github.com/ava-labs/avalanche-network-runner/utils/constants"
+	coreth_params "github.com/ava-labs/coreth/params"
+)
+
+func LoadGenesisMap() (map[string]interface{}, error) {
+	genesis, err := os.ReadFile(constants.LocalGenesisFile)
+	if err != nil {
+		panic(err)
```
felipemadero, line 15: probably not panic here, but return err

utils/chains.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	var genesisMap map[string]interface{}
+	if err = json.Unmarshal(genesis, &genesisMap); err != nil {
+		panic(err)
```
felipemadero, line 19: same

utils/chains.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/ava-labs/avalanche-network-runner/utils/constants"
+	coreth_params "github.com/ava-labs/coreth/params"
+)
+
+func LoadGenesisMap() (map[string]interface{}, error) {
```
felipemadero, line 12: I supposed a function that should return only the new content for the field cChainGenesis, not the full genesis. because it is also beign modified on local after that, I imagine will be best to all genesis updated to be present at local, and this hiddes the first genesis change

network/config.go
```diff
@@ -15,27 +15,20 @@ import (
...
 )
 
 func init() {
-	if err := json.Unmarshal([]byte(defaultCChainConfigStr), &cChainConfig); err != nil {
-		panic(err)
-	}
-	corethCChainGenesis := coreth_params.AvalancheLocalChainConfig
-	cChainGenesisJSON, err := json.Marshal(corethCChainGenesis)
+	var err error
+	cChainConfig, err = utils.LoadGenesisMap()
```
felipemadero, line 28: I expected this to be filled with the cChainConfig, not with all the genesis

local/network.go
```diff
@@ -106,8 +104,6 @@ type localNetwork struct {
 }
 
 var (
-	//go:embed default
-	embeddedDefaultNetworkConfigDir embed.FS
```
felipemadero, line 109: would like for the genesis to be beign loaded here, have all the genesis changes visible at this level, and only have a function in utils to get the new cChainConfig field contents

felipemadero (commented):
probably unit testing is failling due to relative path usage

felipemadero (approved):

PR #316, opened 2022-10-22
Add config file + network upgrade support to blockchain creation

felipemadero:


local/blockchain.go
```diff
@@ -702,49 +691,144 @@ func (ln *localNetwork) reloadVMPlugins(
...
+					return false, fmt.Errorf("couldn't write file at %q: %w", chainConfigPath, err)
+				}
+			}
+		}
+		if chainSpec.NetworkUpgrade != nil {
+			created = true
+			for nodeName := range ln.nodes {
+				nodeRootDir := getNodeDir(ln.rootDir, nodeName)
+				chainConfigDir := filepath.Join(nodeRootDir, chainConfigSubDir)
+				chainUpgradePath := filepath.Join(chainConfigDir, chainAlias, upgradeConfigFileName)
```
holisticode, line 773: Note: this will overwrite a potentially existing file. From the docs:

> Once an upgrade has been activated (a block after the specified timestamp has been accepted), it must always be present in upgrade.json exactly as it was configured at the time of activation (otherwise the node will refuse to start).

I guess most ANR uses will only do one upgrade, but it might be necessary to also test cumulating upgrades, where this might need to be taken into consideration. Might be stuff for a follow-up PR though.

local/blockchain.go
```diff
@@ -702,49 +691,144 @@ func (ln *localNetwork) reloadVMPlugins(
...
+
+		// create config and network upgrade files
+		if chainSpec.ChainConfig != nil {
+			created = true
+			for nodeName := range ln.nodes {
+				nodeRootDir := getNodeDir(ln.rootDir, nodeName)
+				chainConfigDir := filepath.Join(nodeRootDir, chainConfigSubDir)
+				chainConfigPath := filepath.Join(chainConfigDir, chainAlias, configFileName)
+				if err := createFileAndWrite(chainConfigPath, chainSpec.ChainConfig); err != nil {
+					return false, fmt.Errorf("couldn't write file at %q: %w", chainConfigPath, err)
```
holisticode, line 764: Nit: I'd specify in the error message that this is the chain-config file that failed, and in the upgrade-file writing, accordingly.

local/blockchain.go
```diff
@@ -702,49 +691,144 @@ func (ln *localNetwork) reloadVMPlugins(
...
+					return false, fmt.Errorf("couldn't write file at %q: %w", chainUpgradePath, err)
+				}
+			}
+		}
+		// update config info for snapshopt/restart purposes
+		// put into defaults and reset node specifics
+		if chainSpec.ChainConfig != nil {
+			ln.chainConfigFiles[chainAlias] = string(chainSpec.ChainConfig)
+			for nodeName := range ln.nodes {
+				delete(ln.nodes[nodeName].config.ChainConfigFiles, chainAlias)
```
holisticode, line 784: Why do we need this again?

server/server.go
```diff
@@ -410,11 +389,12 @@ func (s *server) waitChAndUpdateClusterInfo(waitMsg string, readyCh chan struct{
 	case <-s.network.stopCh:
 		return
 	case serr := <-s.network.startErrCh:
-		// TODO: decide what to do here, general failure cause network stop()?
-		// maybe try decide if operation was partial (undesired network, fail)
-		// or was not stated (preconditions check, continue)
 		s.log.Warn("async call failed to complete", zap.String("async-call", waitMsg), zap.Error(serr))
-		panic(serr)
+		stopCtx, stopCtxCancel := context.WithTimeout(context.Background(), stopTimeout)
```
holisticode, line 393: Nice. Panic is ugly.

holisticode (commented):
Nice job.

I left a couple of comments which should be easy to address. Only one more important issue.

local/blockchain.go
```diff
@@ -282,39 +293,22 @@ func (ln *localNetwork) setupWalletAndInstallSubnets(
...
 	ctx context.Context,
-	numSubnets uint32,
-	baseWallet primary.Wallet,
-	testKeyAddr ids.ShortID,
+	subnetIDs []ids.ID,
 	pTXs []ids.ID,
-) (primary.Wallet, []ids.ID, error) {
+	clientURI string,
+) (primary.Wallet, error) {
 	println()
```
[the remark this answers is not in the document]
felipemadero: changing all println()

local/blockchain.go
felipemadero: https://github.com/ava-labs/avalanche-network-runner/issues/318

local/blockchain.go
felipemadero: ln.chainConfigFiles[chainAlias] is setting the global network conf, which will be applied to current and added nodes.
as the user requested for the network to use that conf for that blockchain, remove any previous node-only conf related to it, so it will automatically peek the global one.

holisticode (approved):

PR #336, opened 2022-11-04
Add test message implementation + update avalanchego version

StephenButtolph:


dhrubabasu (approved):

local/test_message.go
```diff
@@ -0,0 +1,24 @@
+package local
```
aaronbuchwald, line 1: nit: add license

PR #340, opened 2022-11-08
Async wait

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/302
All server functions that modify network are now mutexed among them. Certain functions are async, just starting the operation (start, create blockchain, create subnet, load snapshpt), but anyway block the other operations. Functions that can be used during network modification are those related to information (most important ones Status, URIs, StreamStatus). Anyway, the results of the functions are not firm until operation end. Health function now is blocking and is the one that enables async wait.


server/server.go
```diff
@@ -527,12 +534,15 @@ func (s *server) CreateBlockchains(ctx context.Context, req *rpcpb.CreateBlockch
 	// to decide cluster/subnet readiness
 	go func() {
 		s.waitChAndUpdateClusterInfo("waiting for custom chains readiness", readyCh, true)
+		s.mu.Unlock()
```
holisticode, line 537: Same dangerous unlock in a go routine?

server/server.go
```diff
@@ -462,6 +465,8 @@ func getNetworkBlockchainSpec(
 }
 
 func (s *server) CreateBlockchains(ctx context.Context, req *rpcpb.CreateBlockchainsRequest) (*rpcpb.CreateBlockchainsResponse, error) {
+	s.mu.Lock()
```
holisticode, line 468: We're having the same pattern here, a lock, then unlock before every return, *except* one guy. Suspicious IMHO...

server/server.go
```diff
@@ -473,49 +478,51 @@ func (s *server) CreateBlockchains(ctx context.Context, req *rpcpb.CreateBlockch
...
 	for _, chainSpec := range chainSpecs {
 		if chainSpec.SubnetId == nil {
-			s.clusterInfo.Healthy = false
+			info.Healthy = false
 		}
 	}
-
-	s.clusterInfo.CustomChainsHealthy = false
+	info.CustomChainsHealthy = false
+	s.setClusterInfo(info)
```
holisticode, line 525: Also, you are setting the info here. Why does setting not have a normal write lock, and instead only a read lock? With a read lock only, some other go routine could be writing to clusterInfo concurrently?

server/server.go
```diff
@@ -527,12 +534,15 @@ func (s *server) CreateBlockchains(ctx context.Context, req *rpcpb.CreateBlockch
...
 		s.waitChAndUpdateClusterInfo("waiting for custom chains readiness", readyCh, true)
+		s.mu.Unlock()
 	}()
 
-	return &rpcpb.CreateBlockchainsResponse{ClusterInfo: s.clusterInfo}, nil
+	return &rpcpb.CreateBlockchainsResponse{ClusterInfo: info}, nil
 }
 
 func (s *server) CreateSubnets(ctx context.Context, req *rpcpb.CreateSubnetsRequest) (*rpcpb.CreateSubnetsResponse, error) {
+	s.mu.Lock()
```
holisticode, line 544: Same pattern again

server/server.go
```diff
@@ -576,14 +587,19 @@ func (s *server) CreateSubnets(ctx context.Context, req *rpcpb.CreateSubnetsRequ
 	// to decide cluster/subnet readiness
 	go func() {
 		s.waitChAndUpdateClusterInfo("waiting for custom chains readiness", readyCh, true)
+		s.mu.Unlock()
```
holisticode, line 590: Please explain these unlocks inside a go routine

server/server.go
```diff
@@ -592,14 +608,12 @@ func (s *server) Health(ctx context.Context, req *rpcpb.HealthRequest) (*rpcpb.H
 		return nil, err
 	}
 
-	s.mu.Lock()
-	defer s.mu.Unlock()
-
-	s.clusterInfo.NodeNames = s.network.nodeNames
-	s.clusterInfo.NodeInfos = s.network.nodeInfos
-	s.clusterInfo.Healthy = true
+	info.NodeNames = s.network.nodeNames
```
holisticode, line 611: Here too, because there's only a read lock, someone could be writing the info at the same time.

server/server.go
```diff
@@ -759,21 +773,23 @@ func (s *server) AddNode(ctx context.Context, req *rpcpb.AddNodeRequest) (*rpcpb
 		return nil, err
 	}
 
-	s.clusterInfo.NodeNames = s.network.nodeNames
-	s.clusterInfo.NodeInfos = s.network.nodeInfos
+	info.NodeNames = s.network.nodeNames
```
holisticode, line 776: Here as well

server/server.go
```diff
@@ -782,21 +798,23 @@ func (s *server) RemoveNode(ctx context.Context, req *rpcpb.RemoveNodeRequest) (
...
```
holisticode, line 801: and here

server/server.go
```diff
@@ -812,30 +830,30 @@ func (s *server) RestartNode(ctx context.Context, req *rpcpb.RestartNodeRequest)
...
```
holisticode, line 833: not repeating again

server/server.go
```diff
@@ -994,28 +1009,29 @@ func (s *server) LoadSnapshot(ctx context.Context, req *rpcpb.LoadSnapshotReques
 	// to decide cluster/subnet readiness
 	go func() {
 		s.waitChAndUpdateClusterInfo("waiting for local cluster readiness", readyCh, true)
+		s.mu.Unlock()
```
holisticode, line 1012: same pattern

server/server.go
```diff
@@ -1049,12 +1065,18 @@ func (s *server) GetSnapshotNames(ctx context.Context, req *rpcpb.GetSnapshotNam
...
-	s.mu.RLock()
+	s.infoMu.RLock()
 	info := s.clusterInfo
-	s.mu.RUnlock()
+	s.infoMu.RUnlock()
 	return info
 }
 
+func (s *server) setClusterInfo(info *rpcpb.ClusterInfo) {
+	s.infoMu.RLock()
```
holisticode, line 1075: In my opinion, if at all, this should be a write lock I think.

What I am understanding here is that you are trying to lock the clusterInfo, but you can't use `s.mu` because that would cause a deadlock. Therefore you introduce a new lock, but that one is going to never lock for writing as far as I can see.

I think it's safe to remove this lock altogether. If I have seen this correctly, every method calling `setClusterInfo` is locked via `s.mu` anyway. Therefore there can't be any write conflict. If this assumption is correct, we might just **assume** that the lock is held here (in this case add the comment).

Otherwise, explain what's going on and why this is needed.

server/server.go
```diff
@@ -576,14 +587,19 @@ func (s *server) CreateSubnets(ctx context.Context, req *rpcpb.CreateSubnetsRequ
...
 	}()
 
-	return &rpcpb.CreateSubnetsResponse{ClusterInfo: s.clusterInfo}, nil
+	return &rpcpb.CreateSubnetsResponse{ClusterInfo: info}, nil
 }
 
 func (s *server) Health(ctx context.Context, req *rpcpb.HealthRequest) (*rpcpb.HealthResponse, error) {
 	s.log.Debug("health")
-	if info := s.getClusterInfo(); info == nil {
+	s.mu.Lock()
```
holisticode, line 598: I am now confused about what `Health` is supposed to be doing. `Health` was at some point just a call which was returning right away, with the information about if the network is healthy or not. 

Now it seems to be blocking, which in my opinion is strange. What if a client wants just to ask if the network is healthy, without waiting? 

Then, we can have a blocking function `WaitUntilHealthy` which is doing just that. That is much clearer in the wording I think. Also, there is no documentation in the code about what these functions are doing. The documentation should say things like if it is blocking, non blocking, if it assumes a lock held etc. Otherwise it's very difficult for users of these functions to understand the code.

server/server.go
```diff
@@ -592,14 +608,12 @@ func (s *server) Health(ctx context.Context, req *rpcpb.HealthRequest) (*rpcpb.H
...
 
-	s.mu.Lock()
-	defer s.mu.Unlock()
-
-	s.clusterInfo.NodeNames = s.network.nodeNames
-	s.clusterInfo.NodeInfos = s.network.nodeInfos
-	s.clusterInfo.Healthy = true
+	info.NodeNames = s.network.nodeNames
+	info.NodeInfos = s.network.nodeInfos
+	info.Healthy = true
```
holisticode, line 613: Also, why does this `Health` function not need to have `CustomChainsHealthy` to have a value? Add this to the documentation.

tests/e2e/e2e_test.go
```diff
@@ -553,14 +521,16 @@ var _ = ginkgo.Describe("[Start/Remove/Restart/Add/Stop]", func() {
 			gomega.Ω(numSubnets).Should(gomega.Equal(0))
 		})
 		ginkgo.By("add 1 subnet", func() {
-			ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
```
holisticode, line 524: Why so high of a timeout now?

tests/e2e/e2e_test.go
```diff
@@ -553,14 +521,16 @@ var _ = ginkgo.Describe("[Start/Remove/Restart/Add/Stop]", func() {
...
 			_, err := cli.CreateSubnets(ctx)
 			cancel()
 			gomega.Ω(err).Should(gomega.BeNil())
 		})
-		ginkgo.By("wait for network to be healthy", func() {
-			// ignore subnet ID here
-			_ = waitForCustomChainsHealthy()
+		ginkgo.By("wait for the network to be healthy", func() {
+			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
```
holisticode, line 530: Why do we need this, if `CreateSubnets` is also blocking?

tests/e2e/e2e_test.go
```diff
@@ -623,27 +587,11 @@ var _ = ginkgo.Describe("[Start/Remove/Restart/Add/Stop]", func() {
...
-					status, err := cli.Status(ctx)
-					cancel()
-					gomega.Ω(err).Should(gomega.BeNil())
-					created = status.ClusterInfo.CustomChainsHealthy
-					if created {
-						continueLoop = false
-					}
-				}
-			}
+			_, err := cli.Health(ctx)
```
holisticode, line 592: I think either there should be an additional `WaitUntilHealthy` method server side, or we should keep this the responsibility of the client to wait until the network is healthy. I am not sure if being this hard on when to return `Health` is appropriate

tests/e2e/e2e_test.go
```diff
@@ -695,30 +643,12 @@ var _ = ginkgo.Describe("[Start/Remove/Restart/Add/Stop]", func() {
...
-				existingSubnetID := status.ClusterInfo.GetSubnets()[0]
-				gomega.Ω(existingSubnetID).Should(gomega.Not(gomega.BeNil()))
-				cancel()
-				return existingSubnetID
-			}
-		}
-	}
+func getExistingSubnetID() string {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	status, err := cli.Status(ctx)
```
holisticode, line 648: This is also weird. I don't think `Status` should be blocking. If you want a Status, you want it right away from the client-side.

Then it's the client's job to wait until a certain Status is reached.

holisticode (commented):
Great ideas. However, introducing locking makes stuff very complex.  I think there are places where we need to review when, and what to lock.

Some functions are now blocking which I don't expect should be blocking from a client-side point of view. They should return right away and have the client decide what to do.

Finally, the changes introduce very important semantic differences to how things were before. There is a lot of documentation missing. I.e. what server side functions do, what they expect and what they return. A function should say if it expects that a lock is held or not when we use locks. Also, that a function is blocking etc.

PR #349, opened 2022-11-11
Support subnet config

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/348


README.md
```diff
@@ -356,13 +356,13 @@ curl -X POST -k http://localhost:8081/v1/control/createblockchains -d '{"pluginD
 avalanche-network-runner control create-blockchains '[{"vm_name":"'$VM_NAME'","genesis":"'$GENESIS_PATH'", "subnet_id": "'$SUBNET_ID'"}]' --plugin-dir $PLUGIN_DIR
 ```
 
-To create a blockchain with a subnet id, and both chain config and network upgrade file paths (requires network restart):
+To create a blockchain with a subnet id, and chain config, network upgrade, subnet config file paths (requires network restart):
```
holisticode, line 359: Nit: when there are multiple elements in a sentence, "and" is usually put before the last one.

cmd/control/control.go
```diff
@@ -172,6 +173,12 @@ func newStartCommand() *cobra.Command {
 		"",
 		"[optional] JSON string of map from chain id to its upgrade file contents",
 	)
+	cmd.PersistentFlags().StringVar(
```
holisticode, line 176: Looks like the flags part and the map unmarshal part  are always the same? Maybe we could put them into functions?

rpcpb/rpc.proto
```diff
@@ -331,6 +337,11 @@ message RestartNodeRequest {
   // If specified, will create a file "chainname/upgrade.json" with
   // the contents provided here.
   map<string, string> upgrade_configs = 5;
+
+  // Map of subnet id to subnet config file contents.
+  // If specified, will create a file "subnetid.json" under subnets config dir with
+  // the contents provided here.
+  map<string, string> subnet_configs = 6;
```
holisticode, line 344: Shouldn't we actually set `optional` to all these configs elements?

rpcpb/rpc.proto
```diff
@@ -359,6 +370,11 @@ message AddNodeRequest {
...
```
holisticode, line 377: As well here

rpcpb/rpc.proto
```diff
@@ -408,6 +424,7 @@ message LoadSnapshotRequest {
   map<string, string> upgrade_configs = 6;
   optional string global_node_config  = 7;
   optional bool reassign_ports_if_used = 8;
+  map<string, string> subnet_configs = 9;
```
holisticode, line 427: ...and here?

holisticode (commented):
All in all looks good to me.

Only what looks like minor issues commented.

rpcpb/rpc.proto
felipemadero: optional makes the value to be nil if not given, which is already the behaviour for maps

rpcpb/rpc.proto
felipemadero: same

rpcpb/rpc.proto
felipemadero: same

rpcpb/rpc.proto
holisticode: OK but to me it's makes things explicit for every one else. I did not know for example  that the value is nil if not given.
 
So I think it's clearer to set it optional where it needs to be optional

cmd/control/control.go
felipemadero: Would rather prefer another PR with changes for all params https://github.com/ava-labs/avalanche-network-runner/issues/352

rpcpb/rpc.proto
felipemadero: also the change should be applied to all map params in RPC so another PR seems best

holisticode (approved):

PR #350, opened 2022-11-15
Fix panics in nil maps when creating blockchains

holisticode:


local/blockchain.go
```diff
@@ -777,12 +777,18 @@ func (ln *localNetwork) createBlockchainConfigFiles(
 		// update config info for snapshopt/restart purposes
 		// put into defaults and reset node specifics
 		if chainSpec.ChainConfig != nil {
+			if ln.chainConfigFiles == nil {
```
felipemadero, line 780: the problem is a nil when creating the network, I would rather init the maps there, to avoid any posible conflict afterwards (the issue belongs to loadSnapshot)

felipemadero:
replace by https://github.com/ava-labs/avalanche-network-runner/pull/351

PR #361, opened 2022-11-23
[Feature] add blockchain_alias blockchain-spec

boulix3:
Hello this Pull Request adds the "blockchain_alias" option in the blockchainSpecs parameter.

It calls the Admin.AliasChain api when the custom blockchain is ready.

With the timestampvm example, using the timestamp alias, you can use 127.0.0.1:9650/ext/bc/timestamp instead of 127.0.0.1:9650/ext/bc/E8isHenre76NMxbJ3munSQatV8GoQ4XKWQg9vD34xMBqEFJGf.


dasconnor (commented):
I think this looks good but want @felipemadero to take a look

local/blockchain.go
```diff
@@ -90,9 +90,42 @@ func (ln *localNetwork) CreateBlockchains(
 	if err := ln.waitForCustomChainsReady(ctx, chainInfos); err != nil {
 		return err
 	}
+
+	ln.RegisterBlockchainAliases(ctx, chainInfos, chainSpecs)
```
felipemadero, line 94: this needs to check for error and return such is that is the case

local/blockchain.go
```diff
@@ -90,9 +90,42 @@ func (ln *localNetwork) CreateBlockchains(
...
+
 	return nil
 }
 
+// if alias is defined in blockchain-specs, registers an alias for the previously created blockchain
+func (ln *localNetwork) RegisterBlockchainAliases(
+	ctx context.Context,
+	chainInfos []blockchainInfo,
+	chainSpecs []network.BlockchainSpec,
+) {
```
felipemadero, line 104: should return error if aliasing operation fails

local/blockchain.go
```diff
@@ -90,9 +90,42 @@ func (ln *localNetwork) CreateBlockchains(
...
+	fmt.Println()
+	ln.log.Info(logging.Blue.Wrap(logging.Bold.Wrap("Registring blockchain aliases...")))
+	for i, chainSpec := range chainSpecs {
+		if chainSpec.BlockchainAlias != nil {
```
felipemadero, line 108: please use this as non optional and compare agains `""`

local/blockchain.go
```diff
@@ -90,9 +90,42 @@ func (ln *localNetwork) CreateBlockchains(
...
 }
 
+// if alias is defined in blockchain-specs, registers an alias for the previously created blockchain
+func (ln *localNetwork) RegisterBlockchainAliases(
+	ctx context.Context,
+	chainInfos []blockchainInfo,
+	chainSpecs []network.BlockchainSpec,
+) {
+	fmt.Println()
+	ln.log.Info(logging.Blue.Wrap(logging.Bold.Wrap("Registring blockchain aliases...")))
```
felipemadero, line 106: Registering

local/blockchain.go
```diff
@@ -90,9 +90,42 @@ func (ln *localNetwork) CreateBlockchains(
...
+	for i, chainSpec := range chainSpecs {
+		if chainSpec.BlockchainAlias != nil {
+			blockchainAlias := *chainSpec.BlockchainAlias
+			chainId := chainInfos[i].blockchainID.String()
+			for nodeName, node := range ln.nodes {
+				ln.log.Info("Registring blockchain alias",
```
felipemadero, line 112: Registering

felipemadero: would prefer only one log for all 5 nodes, telling the specific alias information to set (that is, a log right before the for)

local/blockchain.go
```diff
@@ -90,9 +90,42 @@ func (ln *localNetwork) CreateBlockchains(
...
+					zap.String("alias", blockchainAlias),
+					zap.String("chain-id", chainId),
+					zap.String("node-name", nodeName))
+				err := node.client.AdminAPI().AliasChain(ctx, chainId, blockchainAlias)
+				if err != nil {
+					ln.log.Error("Error while registring blockchain alias",
```
felipemadero, line 118: this error should be returned and not logged

felipemadero (commented):
Left some comments for improvement

local/blockchain.go
```diff
@@ -90,9 +90,42 @@ func (ln *localNetwork) CreateBlockchains(
...
+	for i, chainSpec := range chainSpecs {
+		if chainSpec.BlockchainAlias != nil {
+			blockchainAlias := *chainSpec.BlockchainAlias
+			chainId := chainInfos[i].blockchainID.String()
+			for nodeName, node := range ln.nodes {
+				ln.log.Info("Registring blockchain alias",
+					zap.String("alias", blockchainAlias),
+					zap.String("chain-id", chainId),
+					zap.String("node-name", nodeName))
+				err := node.client.AdminAPI().AliasChain(ctx, chainId, blockchainAlias)
```
holisticode, line 116: Very minor nit: We use the pattern
`if err :=....; err != nil {}` when only an `err` is returned

boulix3:
Awesome, thanks for the feedback.

felipemadero (commented):
LGTM. Will approve once @holisticode nit is addressed

boulix3:
> 
Oh right, I didn't see this one. Fixing right now.

felipemadero (approved):

PR #378, opened 2022-12-16
add linting

danlaine:
Updates the linter settings to be the same as avalanchego uses.

@holisticode @felipemadero 
There are 2 TODOs in here that the linter is complaining about that I'm not sure how to fix. Please have a look at those.

### DIFF
--- a/local/snapshot.go
+++ b/local/snapshot.go
@@ -122,8 +122,8 @@ func (ln *localNetwork) SaveSnapshot(ctx context.Context, snapshotName string) (
 		return "", err
 	}
 	// create main snapshot dirs
-	snapshotDbDir := filepath.Join(snapshotDir, defaultDBSubdir)
-	err = os.MkdirAll(snapshotDbDir, os.ModePerm)
+	snapshotDBDir := filepath.Join(snapshotDir, defaultDBSubdir)
+	err = os.MkdirAll(snapshotDBDir, os.ModePerm)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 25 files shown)


local/snapshot.go
```diff
@@ -122,20 +122,22 @@ func (ln *localNetwork) SaveSnapshot(ctx context.Context, snapshotName string) (
 		return "", err
 	}
 	// create main snapshot dirs
-	snapshotDbDir := filepath.Join(filepath.Join(snapshotDir, defaultDbSubdir))
+	// TODO for CLI team -- what is the line below supposed to do?
```
holisticode, line 125: @felipemadero the first `Join` isn't doing anything, that's why the linter complains. What was the idea?

local/snapshot.go
```diff
@@ -178,8 +180,10 @@ func (ln *localNetwork) loadSnapshot(
 ) error {
 	ln.lock.Lock()
 	defer ln.lock.Unlock()
+	// TODO for CLI team -- what are the lines below supposed to do?
+	// The linter is complaining about them.
 	snapshotDir := filepath.Join(ln.snapshotsDir, snapshotPrefix+snapshotName)
-	snapshotDbDir := filepath.Join(filepath.Join(snapshotDir, defaultDbSubdir))
+	snapshotDbDir := filepath.Join(filepath.Join(snapshotDir, defaultDBSubdir))
```
holisticode, line 186: @felipemadero same here, the first `Join` is not doing anything.

holisticode (changes requested):

local/snapshot.go
felipemadero: ohh this is just an error, just remove the outer join

local/snapshot.go
felipemadero: same

felipemadero (approved):

holisticode (approved):

PR #389, opened 2023-01-05
Revert "Revert "Remove buildir flag reloaded""

StephenButtolph:


local/network.go
```diff
@@ -825,7 +827,8 @@ func (ln *localNetwork) restartNode(
 
 	if binaryPath != "" {
 		nodeConfig.BinaryPath = binaryPath
-		nodeConfig.Flags[config.BuildDirKey] = filepath.Dir(binaryPath)
+		// remove old value if present, to look for vm binaries in new binary path location
+		delete(nodeConfig.Flags, PluginDirKey)
```
StephenButtolph, line 831: This seems pretty jank can't the `Flags` be defined by the user? We should definitely be keeping this override if the user requests it... right?

.github/workflows/build-test-release.yml
```diff
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - main
-    tags:
-      - "*"
```
StephenButtolph, line 8: I don't know why this would be changed in this PR.

StephenButtolph (commented):
@felipemadero Added some questions

.github/workflows/build-test-release.yml
felipemadero: I don't exactly remember. I think we had an issue with the tag name not passing if those lines where included. I believe @dasconnor made the suggestion to me

local/network.go
felipemadero: Yup flags were possibly defined by the user when first created the node. In this case I assumed the user trying to restart a node by specifying a new binary (usually a different avago version in tests), was expecting also for the plugin dir to automatically change. I made this as I realised restart was lacking plugin dir param available. The other and probably cleaner option would be to add that plugin dir param to this function.

.github/workflows/build-test-release.yml
StephenButtolph: My guess here is that it was to avoid goreleaser from performing a release when the tag was made

local/network.go
felipemadero: Added the extra plugin param to RestartNode. Also to AddNode.

.github/workflows/build-test-release.yml
holisticode: But then we can't publish a release by tag? Is this the desired behavior?

local/blockchain.go
```diff
@@ -458,7 +458,7 @@ func (ln *localNetwork) restartNodesWithWhitelistedSubnets(
 		delete(nodeConfig.Flags, config.WhitelistedSubnetsKey)
 
 		ln.log.Info("removing and adding back the node for whitelisted subnets", zap.String("node-name", nodeName))
-		if err := ln.restartNode(ctx, nodeName, "", "", nil, nil, nil); err != nil {
+		if err := ln.restartNode(ctx, nodeName, "", "", "", nil, nil, nil); err != nil {
```
holisticode, line 461: Note, not to be addressed in this PR but possibly create a ticket: These naked parameters are close to impossible to reason about what each means. I think we should not do this?

holisticode (approved):
LGTM, just left a couple of notes for consideration

local/blockchain.go
felipemadero: https://github.com/ava-labs/avalanche-network-runner/issues/390

.github/workflows/build-test-release.yml
felipemadero: PM with @dasconnor give no conclusion on the origin of this haha including it again

server/server.go
```diff
@@ -773,10 +767,12 @@ func (s *server) AddNode(_ context.Context, req *rpcpb.AddNodeRequest) (*rpcpb.A
 		}
 	}
 
+	nodeFlags[PluginDirKey] = req.GetPluginDir()
```
StephenButtolph, line 770: Should this only be populated if `GetPluginDir()` returns a non-empty string?

rpcpb/rpc.proto
```diff
@@ -381,6 +383,9 @@ message AddNodeRequest {
   // If specified, will create a file "subnetid.json" under subnets config dir with
   // the contents provided here.
   map<string, string> subnet_configs = 6;
+
+  // Plugin dir from which to load all custom VM executables.
+  optional string plugin_dir = 7;
```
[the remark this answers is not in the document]
felipemadero: You are so right. This was technical debt. I just copy/paste from Start prototype. Changed so. And added ticket for replacing all:
https://github.com/ava-labs/avalanche-network-runner/issues/391

server/server.go
felipemadero: ok that is more consistent

PR #397, opened 2023-01-10
Fix chain data

patrick-ogrady:


local/network.go
```diff
@@ -962,11 +963,19 @@ func (ln *localNetwork) buildFlags(
 		return buildFlagsReturn{}, err
 	}
 
+	// Use node directory for profiles
+	profileDir := path.Join(nodeDir, "profiles")
```
holisticode, line 967: For file system OS paths, we usually use `filepath`. `path` should be used for URLs only. https://pkg.go.dev/path

holisticode: If this unblocks you in some way, I guess this works. However, essentially this means we're hard-coding these flags to their avago defaults. 

I thought we support the pattern where such things can be set on a global level via custom node configs and/or global node configs, allowing to actually pass individual values for these. 

Curious about @felipemadero 's opinion here. In any case, if this is a blocker, we can work on on a follow up ticket.

patrick-ogrady:
platform recommended we use `--data-dir` instead

felipemadero:
> platform recommended we use `--data-dir` instead

Can we close this? Will prepare a PR following that suggestion

https://github.com/ava-labs/avalanche-network-runner/issues/400

PR #398, opened 2023-01-10
Run e2e test on macos

patrick-ogrady:


.github/workflows/build-test-release.yml
```diff
@@ -40,7 +40,10 @@ jobs:
           CGO_CFLAGS: "-O -D__BLST_PORTABLE__" # Set the CGO flags to use the portable version of BLST
   e2e_test:
     name: e2e tests
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+        matrix:
+            os: [ubuntu-20.04, macos-12]
```
holisticode, line 46: What is the reason you suggest to set to explicit 20.04 vs the current `ubuntu-latest`?

felipemadero (commented):
LGTM after answering @holisticode question. In general do we want to specify ubuntu 20.04 and macos 12? This probably will to be also applied to CLI CI

.github/workflows/build-test-release.yml
felipemadero: ok let's use 20.04 and macos-12, to provide an environment most compatible with avalanchego binaries

felipemadero (approved):

Issue #400, opened 2023-01-10
start using --data-dir flag


felipemadero:

felipemadero:
Closed by https://github.com/ava-labs/avalanche-network-runner/pull/403

PR #403, opened 2023-01-10
Use data dir flag

felipemadero:
This is just a first step on https://github.com/ava-labs/avalanche-network-runner/issues/400 to quickly enable users to have all the node dirs in one separate location, While the bigger update is on the way

### DIFF
--- a/local/network.go
+++ b/local/network.go
@@ -910,7 +910,6 @@ type buildFlagsReturn struct {
 	dbDir     string
 	logsDir   string
 	pluginDir string
-	dataDir   string
 	httpHost  string
 }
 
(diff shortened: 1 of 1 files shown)


local/network.go
```diff
@@ -965,6 +971,7 @@ func (ln *localNetwork) buildFlags(
 	// Flags for AvalancheGo
 	flags := []string{
 		fmt.Sprintf("--%s=%d", config.NetworkNameKey, ln.networkID),
+		fmt.Sprintf("--%s=%s", config.DataDirKey, dataDir),
```
dasconnor, line 974: Bad formatting

dasconnor (approved):
LGTM

local/network.go
felipemadero: which one? don't understand

dasconnor: nvm, this displayed incorrectly on my phone

Issue #414, opened 2023-01-19
Deprecate whitelisted subnets flag


felipemadero:
Instead use new track subnets flag

Issue #415, opened 2023-01-19
Fix deprecated flags when loading snapshots


felipemadero:
This improves loading of snapshots saved with previous ANR versions

felipemadero:
Closed by https://github.com/ava-labs/avalanche-network-runner/pull/416

PR #416, opened 2023-01-19
Use track subnets flag

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/414 https://github.com/ava-labs/avalanche-network-runner/issues/415
It is not changing external RPC + Client usage of whitelisted subnets, only avago flags and internal data structures


client/client.go
```diff
@@ -383,9 +383,15 @@ func WithExecPath(execPath string) OpOption {
 	}
 }
 
-func WithWhitelistedSubnets(whitelistedSubnets string) OpOption {
+func WithWhitelistedSubnets(trackSubnets string) OpOption {
```
holisticode, line 386: Why are we keeping this one?

local/blockchain.go
```diff
@@ -423,41 +423,41 @@ func (ln *localNetwork) getCurrentSubnets(ctx context.Context) ([]ids.ID, error)
...
 // TODO: make this "restart" pattern more generic, so it can be used for "Restart" RPC
-func (ln *localNetwork) restartNodesWithWhitelistedSubnets(
+func (ln *localNetwork) restartNodesWithTrackSubnets(
 	ctx context.Context,
 	subnetIDs []ids.ID,
 ) (err error) {
 	fmt.Println()
-	ln.log.Info(logging.Green.Wrap("restarting each node"), zap.String("whitelisted-subnets", config.WhitelistedSubnetsKey))
-	whitelistedSubnetIDsMap := map[string]struct{}{}
+
```
holisticode, line 431: I suggest to keep this log but in Debug

local/blockchain.go
```diff
@@ -423,41 +423,41 @@ func (ln *localNetwork) getCurrentSubnets(ctx context.Context) ([]ids.ID, error)
...
+	ln.flags[config.TrackSubnetsKey] = trackSubnets
 
 	for nodeName, node := range ln.nodes {
 		// delete node specific flag so as to use default one
 		nodeConfig := node.GetConfig()
-		delete(nodeConfig.Flags, config.WhitelistedSubnetsKey)
+		delete(nodeConfig.Flags, config.TrackSubnetsKey)
 
-		ln.log.Info("removing and adding back the node for whitelisted subnets", zap.String("node-name", nodeName))
+		ln.log.Info("removing and adding back the node for track subnets", zap.String("node-name", nodeName))
```
holisticode, line 460: This could maybe also be a Debug?

scripts/tests.e2e.sh
```diff
@@ -3,15 +3,15 @@ set -e
 
 export RUN_E2E="true"
 # e.g.,
-# ./scripts/tests.e2e.sh 1.9.5 1.9.4 0.4.7
+# ./scripts/tests.e2e.sh 1.9.7 1.9.6 0.4.8
```
holisticode, line 6: Maybe we don't need to always have to update these, as they are just examples. Alternatively move the `DEFAULT_` vars before this block, then we can also put

`# ./scripts/tests.e2e.sh $DEFAULT_VERSION1 $DEFAULT_VERSION2 $DEFAULT_SUBNET_EVM_VERSION` 

Just a suggestion

holisticode (commented):
Basically lgtm, just a couple of Qs

client/client.go
felipemadero: For back compatibility. That is part of the ANR client library API (eg used by CIs). I think we can keep it and plan to remove it in a couple of releases in the future.

local/blockchain.go
felipemadero: this log was kinda useless. it is printing the name of config.WhitelistedSubnetsKey flag

holisticode: It was also printing "restarting each node" which seems useful in Debug. Nevertheless, leave it up to you.

felipemadero: We have a very similar log some lines afterwards, which also prints the tracked ids

holisticode (approved):

PR #420, opened 2023-01-19
Set flags to avago version

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/410


local/network.go
```diff
@@ -1015,3 +1039,67 @@ func (ln *localNetwork) buildFlags(
 		httpHost:  httpHost,
 	}, nil
 }
+
+// Get AvalancheGo version
+func (ln *localNetwork) getNodeSemVer(nodeConfig node.Config) (string, error) {
```
holisticode, line 1044: Could we get this done simpler?

This is what the current binary prints:
`avalanche/1.9.7 [database=v1.4.5, rpcchainvm=22, commit=3e3e40f2f4658183d999807b724245023a13f5dc]`

So maybe a RegEx which just extracts the first string after `/` until the first space, in this case `1.9.7` and then adds `v` in front. Something like this perhaps:

```
match := regexp.MustCompile(`\/(.*?)\S+`)
```

This would reduce this whole block quite a bit.

If you absolutely want to also check that `avalanche` is there (which I think is overkill, as we should be safe to assume that to be the avalanche binary, otherwise we'lrehaving bigger issues), then just match the whole string. 

I also should be able to safely assume that to be a valid semver version https://github.com/ava-labs/avalanchego/blob/master/version/constants.go#L21

local/network.go
```diff
@@ -1015,3 +1052,48 @@ func (ln *localNetwork) buildFlags(
...
+		return "", fmt.Errorf(
+			"invalid version output %q for binary %q: version pattern not found",
+			nodeVersionOutput, nodeConfig.BinaryPath,
+		)
+	}
+	nodeSemVer := "v" + matchs[1]
+	return nodeSemVer, nil
+}
+
+// based on avago version, adapt the given flags to the flag naming and content expected by the version
```
holisticode, line 1077: This explanation sounds much more complicated than it needs to be IMO. Please try something easier, for example `ensure flags are compatible with the running avalanchego version`

Or something like that.

holisticode (approved):
LGTM just please change the documentation of the function a bit, thanks.

Issue #413, opened 2023-01-18
rootDataDir is not always set in `ClusterInfo` returned from ANR


holisticode:
Specifically for the Health call, it seems it's not always set (for example, after loading a snapshot and waiting for health in `pkg/subnet/local.go`.

PR #426, opened 2023-01-26
Async busy wait

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/302 https://github.com/ava-labs/avalanche-network-runner/issues/319 https://github.com/ava-labs/avalanche-network-runner/issues/413 
Changes:
- adds WaitForHealthy, that waits for custom chains healthy status, with 2 min timeout,  returning early on error (with the error). Most useful for: start, create blockchain, create subnet, load snapshot.
- remove unused locks: customChainRestartMu, restartMu
- always set clusterInfo.CustomChainsHealthy to true on Start call
- e2e changes: no more time.Sleep, no busy wait for custom vms healthy instead calling WaitForHealthy, timeouts better defined for the diferent ops (2 mins for Healthy and WaitForHealthy, 15 seconds for the others)
- Status always return current cluster info, even if network has failed


server/server.go
```diff
@@ -404,10 +391,53 @@ func (s *server) waitChAndUpdateClusterInfo(waitMsg string, readyCh chan struct{
...
 		}
+		s.log.Info(fmt.Sprintf("%s ready", msg))
 		s.mu.Unlock()
 	}
 }
 
+// wait until some of this conditions is met:
+// - timeout expires
+// - network operation terminates with error message by setting ErrMsg
+// - network operation termiantes successfully by setting CustomChainsHealthy
```
dasconnor, line 402: typo in `termiantes`

dasconnor (approved):
Looks great! Probably also want @holisticode to take a look

cmd/control/control.go
```diff
@@ -383,6 +384,34 @@ func healthFunc(*cobra.Command, []string) error {
...
+func waitForHealthy(*cobra.Command, []string) error {
+	cli, err := newClient()
+	if err != nil {
+		return err
+	}
+	defer cli.Close()
+
+	ctx, cancel := context.WithTimeout(context.Background(), requestTimeout)
+	resp, err := cli.WaitForHealthy(ctx)
+	cancel()
```
holisticode, line 406: We could use `defer` here too?

server/server.go
```diff
@@ -404,10 +393,40 @@ func (s *server) waitChAndUpdateClusterInfo(waitMsg string, readyCh chan struct{
 			}
 			s.clusterInfo.Subnets = s.network.subnets
 		}
+		s.log.Info(fmt.Sprintf("%s ready", msg))
```
holisticode, line 396: The same way we do with `getClusterInfo`, we could have a `setClusterInfo` which does the locking already, but this can be done in a follow up PR

holisticode (commented):
Nice job! Left a couple of nits

server/server.go
felipemadero: https://github.com/ava-labs/avalanche-network-runner/issues/430

holisticode (approved):

PR #447, opened 2023-02-20
`server` refactor

danlaine:
Significantly changes the implementation of `server/server.go` and `server/network.go`

* Add lock to `server` and `localNetwork` 
* Remove use of channels that I found hard to follow
* Miscellaneous cleanup of names, declarations, etc. 

Still need to do another pass over this and do more testing.


server/network.go
```diff
@@ -358,97 +324,119 @@ func (lc *localNetwork) loadSnapshot(
...
 
-	// node info is already available
 	if err := lc.updateNodeInfo(); err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func (lc *localNetwork) loadSnapshotWait(ctx context.Context, loadSnapshotReadyCh chan struct{}) {
```
danlaine, line 369: replaced by `AwaitHealthy`

server/server.go
```diff
@@ -27,14 +27,41 @@ import (
...
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"go.uber.org/zap"
+	"golang.org/x/exp/maps"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
 )
 
+const (
```
danlaine, line 40: moved up from below

server/network.go
```diff
@@ -358,97 +339,127 @@ func (lc *localNetwork) loadSnapshot(
...
-		nodeInfo := lc.nodeInfos[name]
+	if err := lc.updateNodeInfo(); err != nil {
+		return err
+	}
+
+	if err := lc.updateSubnetInfo(ctx); err != nil {
+		return err
+	}
+
+	// TODO do we need to log this? Seems verbose
```
danlaine, line 432: Open question for you guys

server/server.go
```diff
@@ -404,136 +380,48 @@ func (s *server) waitChAndUpdateClusterInfo(msg string, readyCh chan struct{}, u
...
-			continueLoop = false
-			err = ctx.Err()
+			return &rpcpb.WaitForHealthyResponse{ClusterInfo: s.clusterInfo}, ctx.Err()
 		case <-time.After(1 * time.Second):
 		}
 	}
-	return &rpcpb.WaitForHealthyResponse{ClusterInfo: s.getClusterInfo()}, err
-}
-
-func getNetworkBlockchainSpec(
```
danlaine, line 430: moved to bottom of file

server/network.go
holisticode: Possibly at most Debug log

server/server.go
```diff
@@ -1022,92 +904,98 @@ func (s *server) LoadSnapshot(ctx context.Context, req *rpcpb.LoadSnapshotReques
...
+		defer cancel()
+		err := s.network.AwaitHealthy(ctx)
+		s.mu.Lock()
+		defer s.mu.Unlock()
+		if err != nil {
+			s.log.Warn("snapshot load failed to complete. stopping network and cleaning up network", zap.Error(err))
+
+			s.stopAndRemoveNetwork()
+			return
+		} else {
```
holisticode, line 935: Maybe we don't need an else here as we are return in the != nil case?

server/network.go
danlaine: I'll leave it to you guys to decide -- feel free to push a commit changing this as you see fit

cmd/server/server.go
```diff
@@ -82,31 +85,37 @@ func serverFunc(*cobra.Command, []string) (err error) {
...
-	rootCtx, rootCancel := context.WithCancel(context.Background())
-	errc := make(chan error)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	errChan := make(chan error)
 	go func() {
-		errc <- s.Run(rootCtx)
+		err := s.Run(ctx)
+		if err != nil {
+			log.Error("server Run error", zap.Error(err))
```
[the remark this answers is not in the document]
danlaine: Up to you -- feel free to push a commit changing this

cmd/server/server.go
```diff
@@ -82,31 +85,37 @@ func serverFunc(*cobra.Command, []string) (err error) {
...
 		log.Warn("closed server", zap.Error(waitForServerStop))
-	case serverClosed := <-errc:
-		// server already stopped here
-		_ = rootCancel
+	case serverClosed := <-errChan:
+		// The server stopped.
 		log.Warn("server closed", zap.Error(serverClosed))
 	}
-	return err
+	return nil
```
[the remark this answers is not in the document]
danlaine: This is logged in the above `select`

server/server.go
danlaine: fixed

server/network.go
```diff
@@ -195,7 +184,16 @@ func (lc *localNetwork) createConfig() error {
...
 }
 
-func (lc *localNetwork) start() error {
+// Creates a network and sets [lc.nw] to it.
+// Assumes [lc.lock] isn't held.
+func (lc *localNetwork) Start() error {
+	lc.lock.Lock()
+	defer lc.lock.Unlock()
+
+	if lc.nw != nil {
```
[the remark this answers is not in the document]
danlaine: Same comment as https://github.com/ava-labs/avalanche-network-runner/pull/447#discussion_r1119112807

Would you mind elaborating a bit?

server/server.go
```diff
@@ -27,14 +27,39 @@ import (
...
+	waitForHealthyTimeout = 2 * time.Minute
+
+	rootDataDirPrefix = "network-runner-root-data"
+)
+
+var (
+	ErrInvalidVMName          = errors.New("invalid VM name")
+	ErrInvalidPort            = errors.New("invalid port")
+	ErrNotEnoughNodesForStart = errors.New("not enough nodes specified for start")
+	ErrNetworkRunning         = errors.New("network already running")
```
felipemadero, line 55: I don't think we need to change Err names on err description. This errors where already avaiable with different naming. And they are also part of interface with clients.

server/server.go
```diff
@@ -232,29 +238,28 @@ func (s *server) Ping(context.Context, *rpcpb.PingRequest) (*rpcpb.PingResponse,
...
-	if deadline, ok := ctx.Deadline(); !ok || time.Until(deadline) < defaultStartTimeout {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(context.Background(), defaultStartTimeout)
-		_ = cancel // don't call since "start" is async, "curl" may not specify timeout
-		s.log.Info("received start request with default timeout", zap.String("timeout", defaultStartTimeout.String()))
-	} else {
-		s.log.Info("received start request with existing timeout", zap.String("timeout", deadline.String()))
+	// If [network] is already populated, the network has already been started.
+	if s.network != nil {
+		return nil, ErrNetworkRunning
```
felipemadero, line 247: can use old naming? applies other places also

server/server.go
```diff
@@ -335,67 +321,58 @@ func (s *server) Start(ctx context.Context, req *rpcpb.StartRequest) (*rpcpb.Sta
...
-	readyCh := make(chan struct{})
-	go s.network.startWait(ctx, chainSpecs, readyCh)
-
 	// update cluster info non-blocking
 	// the user is expected to poll this latest information
 	// to decide cluster/subnet readiness
 	go func() {
-		s.waitChAndUpdateClusterInfo("local cluster", readyCh, false)
-		s.waitChAndUpdateClusterInfo("custom chains", readyCh, true)
+		ctx, cancel := context.WithTimeout(context.Background(), waitForHealthyTimeout)
```
felipemadero, line 348: previous semantics was to update info just after network start (before creating chains), and after creating chains. so, almost instantaneously the user can check cluster and node info, without waiting for longer chain creation operation for that

server/server.go
```diff
@@ -232,29 +238,28 @@ func (s *server) Ping(context.Context, *rpcpb.PingRequest) (*rpcpb.PingResponse,
 	return &rpcpb.PingResponse{Pid: int32(os.Getpid())}, nil
 }
 
-const defaultStartTimeout = 5 * time.Minute
+func (s *server) Start(_ context.Context, req *rpcpb.StartRequest) (*rpcpb.StartResponse, error) {
+	s.mu.Lock()
```
felipemadero, line 242: this is a lock over server network/clusterinfo data, another PR was trying to reduce the space of server mutual exclusion by reducing the lock only to parts of code that modify them. This PR is expanding the lock even more

server/server.go
```diff
@@ -335,67 +321,58 @@ func (s *server) Start(ctx context.Context, req *rpcpb.StartRequest) (*rpcpb.Sta
...
-				s.clusterInfo.CustomChains[chainID.String()] = chainInfo.info
-			}
-			s.clusterInfo.Subnets = s.network.subnets
-		}
-		s.log.Info(fmt.Sprintf("%s ready", msg))
-		s.mu.Unlock()
+// Asssumes [s.mu] is held.
+func (s *server) updateClusterInfo() {
+	s.clusterInfo.Healthy = true
+	s.clusterInfo.NodeNames = maps.Keys(s.network.nodeInfos)
```
felipemadero, line 368: this is unsortered

server/server.go
```diff
@@ -404,136 +381,48 @@ func (s *server) waitChAndUpdateClusterInfo(msg string, readyCh chan struct{}, u
...
+		s.mu.RLock()
+
+		if s.network == nil {
+			s.mu.RUnlock()
+			return nil, ErrNetworkNotRunning
+		}
 		if s.clusterInfo.CustomChainsHealthy {
-			break
+			s.mu.RUnlock()
+			return &rpcpb.WaitForHealthyResponse{ClusterInfo: s.clusterInfo}, nil
```
felipemadero, line 397: no lock to access this?

server/server.go
```diff
@@ -404,136 +381,48 @@ func (s *server) waitChAndUpdateClusterInfo(msg string, readyCh chan struct{}, u
...
 		}
+		s.mu.RUnlock()
+
 		select {
-		case err = <-s.asyncErrCh:
-			continueLoop = false
 		case <-ctx.Done():
-			continueLoop = false
-			err = ctx.Err()
+			return &rpcpb.WaitForHealthyResponse{ClusterInfo: s.clusterInfo}, ctx.Err()
```
felipemadero, line 403: same here, no lock?

server/server.go
```diff
@@ -543,143 +432,152 @@ func (s *server) CreateBlockchains(
...
-	if err := s.getNetwork().waitForLocalClusterReady(ctx); err != nil {
+	if err := s.network.AwaitHealthy(ctx); err != nil {
 		return nil, err
 	}
 
-	s.mu.Lock()
-	defer s.mu.Unlock()
-
-	s.clusterInfo.NodeNames = s.network.nodeNames
+	s.clusterInfo.NodeNames = maps.Keys(s.network.nodeInfos)
```
felipemadero, line 528: unsorted

server/server.go
```diff
@@ -543,143 +432,152 @@ func (s *server) CreateBlockchains(
...
-	go s.network.createSubnets(ctx, numSubnets, readyCh)
-
 	// update cluster info non-blocking
 	// the user is expected to poll this latest information
 	// to decide cluster/subnet readiness
 	go func() {
-		s.waitChAndUpdateClusterInfo("custom chains", readyCh, true)
+		ctx, cancel := context.WithTimeout(context.Background(), waitForHealthyTimeout)
+		defer cancel()
+		err := s.network.CreateSubnets(ctx, numSubnets)
```
felipemadero, line 496: we want this error to be returned by WaitForHealthy

server/server.go
```diff
@@ -335,67 +321,58 @@ func (s *server) Start(ctx context.Context, req *rpcpb.StartRequest) (*rpcpb.Sta
...
-
 	// update cluster info non-blocking
 	// the user is expected to poll this latest information
 	// to decide cluster/subnet readiness
 	go func() {
-		s.waitChAndUpdateClusterInfo("local cluster", readyCh, false)
-		s.waitChAndUpdateClusterInfo("custom chains", readyCh, true)
+		ctx, cancel := context.WithTimeout(context.Background(), waitForHealthyTimeout)
+		defer cancel()
+		err := s.network.CreateChains(ctx, chainSpecs)
```
felipemadero, line 350: we want this error to be returned by WaitForHealthy

server/server.go
```diff
@@ -404,136 +381,48 @@ func (s *server) waitChAndUpdateClusterInfo(msg string, readyCh chan struct{}, u
...
+		}
 		if s.clusterInfo.CustomChainsHealthy {
-			break
+			s.mu.RUnlock()
+			return &rpcpb.WaitForHealthyResponse{ClusterInfo: s.clusterInfo}, nil
 		}
+		s.mu.RUnlock()
+
 		select {
-		case err = <-s.asyncErrCh:
```
felipemadero, line 419: we want to notify the cause of async error here (cause for the network to fail)

server/server.go
```diff
@@ -543,143 +432,152 @@ func (s *server) CreateBlockchains(
...
-	go s.network.createBlockchains(ctx, chainSpecs, readyCh)
-
 	// update cluster info non-blocking
 	// the user is expected to poll this latest information
 	// to decide cluster/subnet readiness
 	go func() {
-		s.waitChAndUpdateClusterInfo("custom chains", readyCh, true)
+		ctx, cancel := context.WithTimeout(context.Background(), waitForHealthyTimeout)
+		defer cancel()
+		err := s.network.CreateChains(ctx, chainSpecs)
```
felipemadero, line 454: we want this error to be returned by WaitForHealthy

server/server.go
```diff
@@ -812,50 +712,50 @@ func (s *server) AddNode(_ context.Context, req *rpcpb.AddNodeRequest) (*rpcpb.A
...
 		return nil, err
 	}
 
-	if err := s.network.updateNodeInfo(); err != nil {
+	if err := s.network.UpdateNodeInfo(); err != nil {
 		return nil, err
 	}
 
-	s.clusterInfo.NodeNames = s.network.nodeNames
+	s.clusterInfo.NodeNames = maps.Keys(s.network.nodeInfos)
```
felipemadero, line 743: unsorted

server/server.go
```diff
@@ -869,36 +769,25 @@ func (s *server) RestartNode(ctx context.Context, req *rpcpb.RestartNodeRequest)
...
```
felipemadero, line 776: same

server/server.go
```diff
@@ -869,36 +769,25 @@ func (s *server) RestartNode(ctx context.Context, req *rpcpb.RestartNodeRequest)
...
 
-	info := s.clusterInfo
-	if info == nil {
-		info = &rpcpb.ClusterInfo{}
-	}
-	info.Healthy = false
+	s.stopAndRemoveNetwork()
 
-	return &rpcpb.StopResponse{ClusterInfo: info}, nil
+	return &rpcpb.StopResponse{}, nil
```
felipemadero, line 790: previously this was returning latest valid cluster info

server/server.go
```diff
@@ -1022,92 +903,98 @@ func (s *server) LoadSnapshot(ctx context.Context, req *rpcpb.LoadSnapshotReques
...
-	go s.network.loadSnapshotWait(ctx, readyCh)
-
 	// update cluster info non-blocking
 	// the user is expected to poll this latest information
 	// to decide cluster/subnet readiness
 	go func() {
-		s.waitChAndUpdateClusterInfo("local cluster", readyCh, true)
+		ctx, cancel := context.WithTimeout(context.Background(), waitForHealthyTimeout)
+		defer cancel()
+		err := s.network.AwaitHealthy(ctx)
```
felipemadero, line 926: we want this error to be returned by WaitForHealthy

rpcpb/rpc.proto
```diff
@@ -407,9 +407,7 @@ message AddNodeResponse {
 
 message StopRequest {}
 
-message StopResponse {
```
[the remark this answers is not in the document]
danlaine: The response is always an empty cluster info because the network is stopped

tests/e2e/e2e_test.go
```diff
@@ -627,7 +627,7 @@ var _ = ginkgo.Describe("[Start/Remove/Restart/Add/Stop]", func() {
 			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 			_, err := cli.Health(ctx)
 			cancel()
-			gomega.Ω(err.Error()).Should(gomega.ContainSubstring("not bootstrapped"))
+			gomega.Ω(err.Error()).Should(gomega.ContainSubstring(server.ErrNetworkNotRunning.Error()))
```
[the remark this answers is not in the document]
danlaine: It wasn't accurate -- it was saying "not bootstrapped" but really the error is that the network isn't running at all

tests/e2e/e2e_test.go
```diff
@@ -496,7 +496,7 @@ var _ = ginkgo.Describe("[Start/Remove/Restart/Add/Stop]", func() {
 		})
 		ginkgo.By("calling start API with custom config", func() {
 			ux.Print(log, logging.Green.Wrap("sending 'start' with the valid binary path: %s"), execPath1)
-			ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+			ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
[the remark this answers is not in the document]
danlaine: It was flaking so I bumped it

cmd/server/server.go
```diff
@@ -60,16 +60,19 @@ func serverFunc(*cobra.Command, []string) (err error) {
...
 	}
-	lcfg := logging.Config{
-		DisplayLevel: lvl,
-		LogLevel:     lvl,
-	}
-	lcfg.Directory = logDir
-	logFactory := logging.NewFactory(lcfg)
+
+	logFactory := logging.NewFactory(logging.Config{
+		RotatingWriterConfig: logging.RotatingWriterConfig{
```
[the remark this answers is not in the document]
danlaine: This is semantically the same as what was here before

cmd/server/server.go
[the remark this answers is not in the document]
danlaine: I'll defer to you guys on logging -- whatever you want

server/network.go
```diff
@@ -20,10 +19,12 @@ import (
 	"github.com/ava-labs/avalanchego/ids"
 	avago_constants "github.com/ava-labs/avalanchego/utils/constants"
 	"github.com/ava-labs/avalanchego/utils/logging"
+	"golang.org/x/exp/maps"
 )
 
 type localNetwork struct {
-	log logging.Logger
+	lock sync.Mutex
```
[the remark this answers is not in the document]
danlaine: We start `localNetwork` methods in a goroutine in some places in `server`, so we need a lock to ensure we're not concurrently accessing the state of `localNetwork`

server/network.go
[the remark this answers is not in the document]
danlaine: This check might not be _strictly_ necessary but it makes the code pattern nicer -- namely, exported methods use a lock

server/network.go
```diff
@@ -215,122 +213,105 @@ func (lc *localNetwork) start() error {
...
-	var ctx context.Context
-	ctx, lc.startCtxCancel = context.WithCancel(argCtx)
+) error {
+	lc.lock.Lock()
+	defer lc.lock.Unlock()
 
-	if err := lc.waitForLocalClusterReady(ctx); err != nil {
-		lc.startErrCh <- err
-		return
+	if lc.nw == nil {
```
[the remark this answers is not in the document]
danlaine: https://github.com/ava-labs/avalanche-network-runner/pull/447#discussion_r1120171717

^ Same as this comment

server/network.go
```diff
@@ -215,122 +213,105 @@ func (lc *localNetwork) start() error {
...
+			cancel()
+		case <-ctx.Done():
+			// This method is done. Don't leak [ctx].
+		}
+	}(ctx)
 
-	if err := lc.updateNodeInfo(); err != nil {
-		lc.startErrCh <- err
-		return
+	if err := lc.awaitHealthy(ctx); err != nil {
```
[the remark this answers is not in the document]
danlaine: I'll leave the naming up to you all

server/network.go
```diff
@@ -195,7 +184,16 @@ func (lc *localNetwork) createConfig() error {
 	return nil
 }
 
-func (lc *localNetwork) start() error {
+// Creates a network and sets [lc.nw] to it.
+// Assumes [lc.lock] isn't held.
+func (lc *localNetwork) Start() error {
```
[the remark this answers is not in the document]
danlaine: Exported (capitalized) methods are supposed to be used by external callers and grab a lock, non-exported methods aren't and don't grab a lock.

server/network.go
```diff
@@ -358,97 +339,127 @@ func (lc *localNetwork) loadSnapshot(
...
-}
-
+// Populates [lc.customChainIDToInfo] for all chains other than those on
+// the Primary Network (P-Chain, X-Chain, C-Chain.)
+// Populates [lc.subnets] with all subnets that exist.
+// Doesn't contain the Primary network.
+// Assumes [lc.lock] is held.
 func (lc *localNetwork) updateSubnetInfo(ctx context.Context) error {
-	node, err := lc.nw.GetNode(lc.nodeNames[0])
+	nodeNames := maps.Keys(lc.nodeInfos)
```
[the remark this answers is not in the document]
danlaine: Does it matter which node we query this info from?

server/server.go
danlaine: I don't think we had errors for "network already running" or "no network is running"

server/server.go
danlaine: The old naming was misleading -- it said "not bootstrapped" but it really meant "not running"

server/server.go
danlaine: We update `clusterInfo` before starting the network:

```go
	s.clusterInfo = &rpcpb.ClusterInfo{
		Pid:         pid,
		RootDataDir: rootDataDir,
	}
```

And also after:

```go
	go func() {
		ctx, cancel := context.WithTimeout(context.Background(), waitForHealthyTimeout)
		defer cancel()
		err := s.network.CreateChains(ctx, chainSpecs)
		s.mu.Lock()
		defer s.mu.Unlock()
		if err != nil {
			s.log.Error("network never became healthy", zap.Error(err))
			s.stopAndRemoveNetwork()
			return
		}
		s.updateClusterInfo()
		s.log.Info("network healthy")
	}()
```

server/server.go
danlaine: > This PR is expanding the lock even more

I think this is what it takes in order for `server` to be threadsafe -- we can't have multiple goroutines simultaneously accessing the state of `server` or else it's unsafe

server/server.go
danlaine: https://github.com/ava-labs/avalanche-network-runner/pull/447#discussion_r1120195954

server/server.go
danlaine: I don't understand -- this channel doesn't exist anymore. Where are the unhandled errors?

server/server.go
danlaine: https://github.com/ava-labs/avalanche-network-runner/pull/447#discussion_r1120195954

server/server.go
danlaine: The clusterInfo will always be empty so I don't see a need to return it

server/server.go
```diff
@@ -968,45 +866,28 @@ func (s *server) SendOutboundMessage(ctx context.Context, req *rpcpb.SendOutboun
...
-	} else {
-		s.log.Info("received start request with existing timeout", zap.String("timeout", deadline.String()))
-	}
-
-	// If [network] is already populated, the network has already been started.
-	if s.getNetwork() != nil {
-		return nil, ErrAlreadyBootstrapped
-	}
-
+func (s *server) LoadSnapshot(_ context.Context, req *rpcpb.LoadSnapshotRequest) (*rpcpb.LoadSnapshotResponse, error) {
```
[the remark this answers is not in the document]
danlaine: It isn't used in this method -- this method is non-blocking

server/server.go
danlaine: Not sure what you mean

server/server.go
danlaine: sorted now

server/server.go
danlaine: sorted now

server/server.go
danlaine: fixed

server/server.go
danlaine: fixed

server/server.go
danlaine: sorted now

server/server.go
danlaine: https://github.com/ava-labs/avalanche-network-runner/pull/447#discussion_r1120182965

server/network.go
```diff
@@ -358,97 +339,127 @@ func (lc *localNetwork) loadSnapshot(
...
+	if err := lc.updateNodeInfo(); err != nil {
+		return err
+	}
+
+	if err := lc.updateSubnetInfo(ctx); err != nil {
+		return err
+	}
+
+	// TODO do we need to log this? Seems verbose
+	for name, nodeInfo := range lc.nodeInfos {
```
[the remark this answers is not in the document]
danlaine: Sorted now

server/network.go
```diff
@@ -358,97 +339,127 @@ func (lc *localNetwork) loadSnapshot(
...
 	lc.subnets = []string{}
 	for _, subnet := range subnets {
 		if subnet.ID != avago_constants.PlatformChainID {
 			lc.subnets = append(lc.subnets, subnet.ID.String())
 		}
 	}
-	for _, nodeName := range lc.nodeNames {
-		nodeInfo := lc.nodeInfos[nodeName]
+
+	for _, nodeInfo := range lc.nodeInfos {
```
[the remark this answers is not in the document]
danlaine: sorted now

cmd/server/server.go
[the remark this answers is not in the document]
felipemadero: pushing a commit for changing this

rpcpb/rpc.proto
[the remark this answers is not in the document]
felipemadero: I sugest you to check the output of the command `go run main.go control stop`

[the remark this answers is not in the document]
felipemadero: I would like to preserve actual ANR behaviour whenever posible

tests/e2e/e2e_test.go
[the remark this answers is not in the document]
felipemadero: Could we keep the original naming? This will propagate into CLI and represents and API change, would like to include it in a bigger refactor related to API changes with appropiate semver change.

server/network.go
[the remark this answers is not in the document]
felipemadero: let's just remove the check

server/network.go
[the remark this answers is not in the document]
felipemadero: let's remove this

server/network.go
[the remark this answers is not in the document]
felipemadero: Yup it is best for debug to know which node is interacting ANR with

server/server.go
felipemadero: we had ErrAlreadyBootstrapped and ErrNotBootstrapped with same semantics. Would prefer to keep them until bigger change to API.

server/server.go
felipemadero: the previous situation was for clusterInfo to be updated after starting the network, with the node names etc, and secondly after creating the blockchains, with the subnet/blockchain infoi

felipemadero: anyway we can keep this, it is a minor semantic change not noticeable for users

server/server.go
felipemadero: WaitForHealthy not only waits for network to be on, it is also giving to the client the reason for the network failure if there was any. Usually that error was hidden in ANR logs, recently we added this function so users like CLI can check the error and provide it to the user.
All the async errors from go routines, due to calling CreateChains,  CreateSubnets, AwaitAndUpdateNetworkInfo, should be passed to this function using some mechanism.

server/server.go
felipemadero: https://github.com/ava-labs/avalanche-network-runner/pull/447#discussion_r1131592992

server/server.go
felipemadero: clusterinfo was being returned on stop, pls try to keep the same semantics

server/server.go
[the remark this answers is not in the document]
felipemadero: ok... I understood that maybe the ANR client will want to provide a timeout on the async op. is that reasonable?

server/server.go
felipemadero: https://github.com/ava-labs/avalanche-network-runner/pull/447#discussion_r1131592992

server/server.go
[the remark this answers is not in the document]
danlaine: We can't use the context passed into `LoadSnapshot` for the async operations in the goroutine started by this method because the `ctx` will probably be cancelled after `LoadSnapshot` returns, which will kill the async goroutine.

Namely, the caller is probably doing something like:

```go
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
result, err := server.LoadSnapshot(ctx, req)
cancel() // This kills the async goroutine even though we didn't mean to
```

dasconnor (approved):
LGTM, great job everyone!

PR #452, opened 2023-02-20
nits

danlaine:
no functional changes, just nits


local/blockchain.go
```diff
@@ -683,6 +677,7 @@ func (ln *localNetwork) waitSubnetValidators(
 				nodeID := node.GetNodeID()
 				if isValidator := subnetValidators.Contains(nodeID); !isValidator {
 					ready = false
+					break subnetLoop
```
danlaine, line 680: I'm assuming we do want to break here?

local/network.go
```diff
@@ -350,22 +352,6 @@ func NewDefaultNetwork(
...
-	return outFlags
-}
-
-func copyMapStringString(flags map[string]string) map[string]string {
-	outFlags := map[string]string{}
-	for k, v := range flags {
-		outFlags[k] = v
-	}
-	return outFlags
-}
```
danlaine, line 367: Replaced by `maps.Clone`

local/snapshot.go
```diff
@@ -116,11 +117,7 @@ func (ln *localNetwork) SaveSnapshot(ctx context.Context, snapshotName string) (
 		nodeConfig := node.config
 		// depending on how the user generated the config, different nodes config flags
 		// may point to the same map, so we made a copy to avoid always modifying the same value
-		nodeConfigFlags := make(map[string]interface{})
-		for fk, fv := range nodeConfig.Flags {
-			nodeConfigFlags[fk] = fv
-		}
-		nodeConfig.Flags = nodeConfigFlags
+		nodeConfig.Flags = maps.Clone(nodeConfig.Flags)
```
danlaine, line 120: So, this change is functionally equivalent to what was here before, but it's actually a no-op. Can we remove or fix this?

local/blockchain.go
felipemadero: we want to wait for waitForValidatorsPullFrequency

local/snapshot.go
felipemadero: It is not a no-op. It is creating a new map. To solve situation where different nodes share the same map. Probably will require a fix but I think not here.

local/snapshot.go
```diff
@@ -180,16 +174,13 @@ func (ln *localNetwork) SaveSnapshot(ctx context.Context, snapshotName string) (
 		SubnetConfigFiles:  ln.subnetConfigFiles,
 	}
 
-	for _, nodeConfig := range nodesConfig {
-		// no need to save this, will be generated automatically on snapshot load
-		networkConfig.NodeConfigs = append(networkConfig.NodeConfigs, nodeConfig)
-	}
+	// no need to save this, will be generated automatically on snapshot load
+	networkConfig.NodeConfigs = append(networkConfig.NodeConfigs, maps.Values(nodesConfig)...)
```
felipemadero, line 178: why not just nodesConfig.... instead of maps.Values(nodesConfig)... ? I think the original copies the same object?

local/blockchain.go
```diff
@@ -104,19 +105,20 @@ func (ln *localNetwork) RegisterBlockchainAliases(
 	chainInfos []blockchainInfo,
 	chainSpecs []network.BlockchainSpec,
 ) error {
-	fmt.Println()
+	fmt.Println() // TODO can we remove all these empty line prints?
```
felipemadero, line 108: this is old heritage. improves a little the visual. I am not opposite. What do you think @dasconnor @holisticode ?

local/blockchain.go
```diff
@@ -668,6 +663,7 @@ func (ln *localNetwork) waitSubnetValidators(
 	ln.log.Info(logging.Green.Wrap("waiting for the nodes to become subnet validators"))
 	for {
 		ready := true
+	subnetLoop:
```
felipemadero, line 666: would prefer not to use labels in general

felipemadero (commented):
I also added some comments from outside this review

local/blockchain.go
holisticode: A run produces a lot of logs. It is often pretty difficult to find things. I like having some visual separation of things. 

But then, I usually need to look for logs in the log file, and these lines aren't put into the log files? Therefore I am indifferent to this issue. Whatever you prefer.

local/blockchain.go
danlaine: With this change, we will still do that. Without this `break` we will do a bunch of useless work -- namely we will do the logic inside this loop:

```go
for _, subnetID := range subnetIDs {
```

even though the logic inside that loop doesn't matter because `ready` is already false.

local/blockchain.go
danlaine: Personally I'd prefer to remove but up to you guys. Let me know either way and I'll either remove the TODO or the `fmt.Println()`

local/snapshot.go
danlaine: `To solve situation where different nodes share the same map.` 

Ok if that's the case then it's not a no-op. This change is semantically the same as before so should be fine.

local/snapshot.go
danlaine: This is functionally equivalent to what was here before. We can't append `nodesConfig` because it's a `map[string]node.Config`, not a `node.Config`. If you look inside the implementation of `maps.Values` you'll see it's doing the exact same thing:

```go
// Values returns the values of the map m.
// The values will be in an indeterminate order.
func Values[M ~map[K]V, K comparable, V any](m M) []V {
	r := make([]V, 0, len(m))
	for _, v := range m {
		r = append(r, v)
	}
	return r
}
```

local/blockchain.go
danlaine: Yeah, me too, but it's the only way to avoid doing unnecessary loop iterations

local/blockchain.go
dasconnor: I'm on the fence as well, I'd say let's just leave them.

dasconnor (approved):
Let's leave the prints. Other than that, I think everything else is resolved?

local/blockchain.go
felipemadero: You can use an `if !ready` and a simple `break` as the last sentence in the outer for

local/blockchain.go
felipemadero: Pls remove the TODO

local/blockchain.go
danlaine: done

local/blockchain.go
danlaine: done

felipemadero (approved):

Issue #436, opened 2023-02-08
add RPC protocol version management


felipemadero:
- server provide a RPC protocol version in a API method
- client knows its RPC protocol version and denies to connect to another version of the server

PR #457, opened 2023-02-24
Add rpc protocol version

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/436


server/server.go
```diff
@@ -88,6 +88,9 @@ var (
 )
 
 const (
+	// RPCVersion should be bumped anytime changes are made which require
+	// the RPC client to upgrade to latest RPC server to be compatible
+	RPCVersion   string = "1.0.0"
```
dasconnor, line 93: I think you need to set this in your proto file, don't you?

dasconnor: Also, Avago uses an integer convention, not semantic version

felipemadero: Constants are not supported in proto 3 AFAIK

dasconnor (approved):

PR #463, opened 2023-03-06
Register BLS Keys + Other Cleanup

patrick-ogrady:
* Register BLS Keys when adding new validators
* Only restart nodes that need to be restarted when adding Subnets/modifying configs
* Very inconsistent which functions block vs return right away
* keep same backend across wallet restarts (to avoid duplicate UTXO usage)


local/blockchain.go
```diff
@@ -205,12 +208,40 @@ func (ln *localNetwork) installCustomChains(
...
+				for nodeName, node := range ln.nodes {
+					if node.nodeID == val.NodeID {
+						nodesToRestart[nodeName] = struct{}{}
+						break
+					}
+				}
+			}
+		}
+
+		for nodeName := range nodesToRestart {
```
patrick-ogrady, line 234: try to find a way to only do this restart once (when tracking new subnets)

network/network.go
```diff
@@ -64,5 +64,5 @@ type Network interface {
 	// Create the specified blockchains
 	CreateBlockchains(context.Context, []BlockchainSpec) error
 	// Create the given numbers of subnets
-	CreateSubnets(context.Context, uint32) error
+	CreateSubnets(context.Context, uint32, []string) error
```
patrick-ogrady, line 67: Thinking about just creating a separate function to have more control over this without modifying existing logic.

local/blockchain.go
```diff
@@ -534,28 +595,75 @@ func (ln *localNetwork) addPrimaryValidators(
...
 		)
 		cancel()
 		if err != nil {
 			return err
 		}
 		ln.log.Info("added node as primary subnet validator", zap.String("node-name", nodeName), zap.String("node-ID", nodeID.String()), zap.String("tx-ID", txID.String()))
+		addedValidators = append(addedValidators, nodeID)
+	}
+
+	// Ensure all added nodes are primary validators (else add subnet validator will revert)
```
patrick-ogrady, line 644: This is needed to avoid the `not committed` error

PR #466, opened 2023-03-06
Simulated wallet

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/482


local/blockchain.go
```diff
@@ -469,10 +475,42 @@ func (ln *localNetwork) restartNodesWithTrackSubnets(
 	return nil
 }
 
+func setupSimulatedPWallet(
```
patrick-ogrady, line 478: So, the reason we run into this issue is that we reset the wallet (which already tracks pending) and when restarting it doesn't know about pending txs?

patrick-ogrady: I wonder if we could just keep the `pBackend` in the local network struct so we could preserve this across restarts more easily? Then, I don't think we'd need to do this tx tracking.

felipemadero: The reason for this issue is that we were building two consecutive blockchain creation txs without having the wallet keep track of consumed UTXOs of the first blockchain create tx build. The goal of the code is to generate a bunch of txs without issuing them to the platform chain yet.
For that to happend. wallet.IssueTx is not called. But due to that backend.AcceptTx is not called, and so, the UTXOs are not tracked as consumed.
Previously the sequence was: `Builder().NewCreateChainTx` + `Signer().SignUnsigned` + `Builder().NewCreateChainTx` + `Signer().SignUnsigned`  for the 2 chains.
Now the sequence is: `Builder().NewCreateChainTx` + `Signer().SignUnsigned` + `Backend.AcceptTx` +  `Builder().NewCreateChainTx` + `Signer().SignUnsigned`  + `Backend.AcceptTx`

felipemadero: I did not found a way in the wallet to generate a sequence of txs with UTXOs and TXs tracking, without issuing them to platform. Probably it is there but I did not see it.
If not there, would like to make a PR to avago to enable this "simulation". Add an option to the wallet so `pClient.IssueTx` is not called when calling to `wallet.IssueTx`

felipemadero: Regarding tx tracking, we need to initialize the wallet with previously created subnets, but we surely can find a way to keep tracking the subnets created during the function execution. What we need is a way to modify the pchain client API is that is changed on restart (port). Maybe we can add something to the wallet interface to do that change without restarting the wallet.

felipemadero: What do you mean by "pending txs"?

felipemadero: Besides previous comments. probably I still need a simulated wallet, and it will simulate UTXO consumption, and later on I will need a non simulated wallet to really issue the txs, and that one will need to not have those UTXOs consumed.

PR #467, opened 2023-03-07
CreateSpecificBlockchains

patrick-ogrady:
Replaces: #463 
* Register BLS Keys when adding new validators
* Only restart nodes/track specific nodes
* keep same backend across wallet restarts (to avoid duplicate UTXO usage)


local/blockchain.go
```diff
@@ -98,6 +98,29 @@ func (ln *localNetwork) CreateBlockchains(
 	return nil
 }
 
+func (ln *localNetwork) CreateSpecificBlockchains(
```
patrick-ogrady, line 101: TODO

server/network.go
```diff
@@ -282,6 +282,51 @@ func (lc *localNetwork) createBlockchains(
 	close(createBlockchainsReadyCh)
 }
 
+func (lc *localNetwork) createSpecificBlockchains(
```
patrick-ogrady, line 285: TODO

patrick-ogrady:
Super close: 
image

server/server.go
```diff
@@ -592,6 +595,78 @@ func (s *server) CreateBlockchains(
...
+		if chainSpec.SubnetID != nil {
+			return nil, ErrSubnetNotNil
+		}
+		chainSpecs = append(chainSpecs, chainSpec)
+	}
+	if len(chainSpecs) == 0 {
+		return nil, ErrNoBlockchainSpec
+	}
+
+	chains, err := s.network.createSpecificBlockchains(ctx, req.ExecPath, chainSpecs, s.cfg.RedirectNodesOutput)
```
felipemadero, line 643: Always have preferred all of this to be blocking, but grpc server functions were async since start. I think that was to try to easy certain client apps. But usually our client apps just prefer to block. Anyway, we have the WaitForHealthy now, that also returns the async error. Maybe to be uniform, this new one should also be non blocking and force the user to also call WaitForHealthy?

local/experimental/blockchain.go
```diff
@@ -0,0 +1,490 @@
...
+		addedValidators = append(addedValidators, nodeID)
+	}
+
+	// Ensure all added nodes are primary validators (else add subnet validator will revert)
+	for ctx.Err() == nil {
+		vs, err := w.client.GetCurrentValidators(ctx, ids.Empty, nil)
+		if err != nil {
+			return err
+		}
+		curValidators := make(map[ids.NodeID]struct{})
```
felipemadero, line 193: maybe a set can be used here

local/experimental/blockchain.go
```diff
@@ -0,0 +1,490 @@
...
+		)
+		cancel()
+		if err != nil {
+			return err
+		}
+		w.log.Info("added node as primary network validator", zap.String("name", nodeName), zap.String("nodeID", nodeID.String()), zap.String("txID", txID.String()))
+		addedValidators = append(addedValidators, nodeID)
+	}
+
+	// Ensure all added nodes are primary validators (else add subnet validator will revert)
```
felipemadero, line 187: I am not sure avoid this wait, maybe the network want to do something else before the nodes are validators, this wait is about min 20 secs I think?

local/experimental/blockchain.go
```diff
@@ -0,0 +1,490 @@
...
+	GetNodeUnsafe(name string) node.Node
+	Logger() logging.Logger
+	SetBlockchainConfigFilesUnsafe(chainSpecs []network.BlockchainSpec, blockchainTxs []*txs.Tx) bool
+	RestartNodeUnsafe(ctx context.Context, nodeName string) error
+	HealthyUnsafe(ctx context.Context) error
+	ReloadVMPluginsUnsafe(ctx context.Context) error
+	AddNodeUnsafe(nodeConfig node.Config) error
+}
+
+type wallet struct {
```
felipemadero, line 51: this surely will be the first thing added to local, in the other PR

local/experimental/blockchain.go
```diff
@@ -0,0 +1,490 @@
...
+	addr   ids.ShortID
+	log    logging.Logger
+
+	client  platformvm.Client
+	backend p.Backend
+	builder p.Builder
+	signer  p.Signer
+}
+
+func newWallet(ctx context.Context, uri string, logger logging.Logger) (*wallet, error) {
```
felipemadero, line 62: but we need Txs for the case subnet ids are given

local/experimental/blockchain.go
felipemadero: it is a p wallet probably can be called pWallet

local/experimental/blockchain.go
felipemadero: newPWallet

local/experimental/blockchain.go
```diff
@@ -0,0 +1,490 @@
...
+		ready := true
+		for _, nodeID := range addedValidators {
+			if _, ok := curValidators[nodeID]; !ok {
+				ready = false
+			}
+		}
+		if ready {
+			break
+		}
+		time.Sleep(1 * time.Second)
```
felipemadero, line 206: maybe we can use a select and time.After

local/experimental/blockchain.go
```diff
@@ -0,0 +1,490 @@
...
+	for _, v := range vs {
+		primaryValidatorsEndtime[v.NodeID] = time.Unix(int64(v.EndTime), 0)
+	}
+	for _, spec := range chainSpecs {
+		subnetID, err := ids.FromString(*spec.SubnetID)
+		if err != nil {
+			return err
+		}
+		nodeIDs := make([]ids.NodeID, len(spec.Participants))
+		for i, nodeName := range spec.Participants {
```
felipemadero, line 348: maybe we can have a wallet function to add subnet validators also

local/experimental/blockchain.go
```diff
@@ -0,0 +1,490 @@
...
+			}
+			w.log.Info("added node as a subnet validator",
+				zap.String("name", nodeName),
+				zap.String("nodeID", nodeID.String()),
+				zap.String("subnetID", subnetID.String()),
+				zap.String("txID", txID.String()),
+			)
+		}
+
+		// Wait for validators to become active
```
felipemadero, line 379: also maybe this can be split, as a network may want to do something in the middle

local/experimental/blockchain.go
```diff
@@ -0,0 +1,490 @@
...
+			ready := true
+			for _, nodeID := range nodeIDs {
+				if _, ok := curValidators[nodeID]; !ok {
+					ready = false
+				}
+			}
+			if ready {
+				break
+			}
+			time.Sleep(1 * time.Second)
```
felipemadero, line 398: select here?

local/experimental/blockchain.go
```diff
@@ -0,0 +1,490 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	wallet.reload(clientURI)
+
+	// Create blockchains
+	// if err := ln.ReloadVMPluginsUnsafe(ctx); err != nil {
+	// 	return nil, nil, err
+	// }
+	if err := wallet.createBlockchains(ctx, chainSpecs, blockchainTxs); err != nil {
```
felipemadero, line 486: it is nice to see the same wallet can be used to issue the txs after it was also used to create those same tx with UTXOs and tx tracking

felipemadero (commented):
I am ok to merge this, with comments addressed. Probably with all names changed to FoorBarExperimental. Probably also keeping server async, unless we want that change in server behaviour.
And as soon as posible, we may want to migrate this to local lib, following the style there. Probably overload CreateSubnets a little more with the funcitonality. Also avoid server to block, unless we will want server to be sync now. I am also tempted to enable AddNode to create multiple nodes, and to be more flexible regarding the options (not to ask for a full config). But this is the least idea.

server/utils.go
```diff
@@ -0,0 +1,19 @@
+package server
+
+import (
+	"encoding/json"
+
+	"github.com/ava-labs/avalanche-network-runner/rpcpb"
+)
+
+func deepCopy(i *rpcpb.ClusterInfo) (*rpcpb.ClusterInfo, error) {
```
patrick-ogrady, line 9: Testing this now in HyperSDK: https://github.com/ava-labs/hypersdk/pull/95

patrick-ogrady: It may only be necessary to clone the strings in the `chains` field in `CreateSpecificBlockchainsResponse`.

Still researching why any of this would be related (assuming this actually solves the issue)

patrick-ogrady: It is possible that we may just need to escape \n properly?

patrick-ogrady: It is also possible that there is a race such that `clusterInfo` changes while it is being encoded?

server/server.go
```diff
@@ -662,9 +675,18 @@ func (s *server) CreateSpecificBlockchains(
...
+
+	// Deep copy struct to avoid shared pointers in
+	// gRPC marshaling (which can trigger race conditions and message decoding
+	// failure)
+	clusterInfo, err := deepCopy(s.clusterInfo)
+	if err != nil {
+		return nil, err
 	}
-	return &rpcpb.CreateSpecificBlockchainsResponse{ClusterInfo: s.clusterInfo, Chains: chainIDs}, nil
+	return &rpcpb.CreateSpecificBlockchainsResponse{ClusterInfo: clusterInfo, Chains: chainIDs}, nil
```
patrick-ogrady, line 689: Technically, returning `s.clusterInfo` in this response is not safe as it may be modified as soon as the lock is released.

felipemadero:
Closing. Pretty important PR with lots of new implementations! All of them where added to local/blockchain.go on v1.4.0. Thanks!

PR #503, opened 2023-04-10
Have Subnet Participants Map as part of Cluster Object

### LINKED ISSUE #501: add subnet id to subnet participants map to cluster info
should be filled in updateSubnetInfo

sukantoraymond:
Closes #501


server/network.go
```diff
@@ -46,6 +46,8 @@ type localNetwork struct {
 	stopOnce sync.Once
 
 	subnets []string
+	// map from subnet ID to participating node info
+	subnetParticipants map[string][]*rpcpb.NodeInfo
```
felipemadero, line 50: would rather map subnet ID to node name, so as to not repeat node info in status, and let the user find out more information on the node

server/network.go
```diff
@@ -377,6 +380,14 @@ func (lc *localNetwork) updateSubnetInfo(ctx context.Context) error {
 		}
 	}
 
+	for subnetID, nodeNames := range lc.nw.GetSubnetParticipants() {
```
felipemadero, line 383: I would not extende ANR local for this, but use pchain gesubnets and getcurrentvalidators to get the subnet participants. this is also more suitable in order to rehuse network saved on snapshots. will work without problem in that case. as a general rule: when possible, just try to get information from pchain (from the blockchain), so things like snapshots does not need to save more information that needed

rpcpb/rpc.proto
```diff
@@ -170,6 +170,10 @@ service ControlService {
   }
 }
 
+message SubnetParticipants {
+  repeated NodeInfo subnet_participants = 1;
```
felipemadero, line 174: I would rather prefer a list of strings, with the node names

felipemadero (commented):
I would rather prefer to not add participants bookeeping to ANR local lib, but rather take the info from the blockchain itself

PR #509, opened 2023-04-11
Wrap P Chain TX error

### LINKED ISSUE #500: wrap all errors related to pWallet tx issuance so as to give them more context
currently only a general uncommitted msg is obtained. wrap all calls with fmt.Errorf("CALL CTX %w", err).
For example at w.pWallet.IssueAddPermissionlessValidatorTx. Context should be name of function at least.
better if including some general arg like blockchain id, node id, whatever main arg for the call

sukantoraymond:
closes #500

### DIFF
--- a/local/blockchain.go
+++ b/local/blockchain.go
@@ -638,7 +638,7 @@ func (ln *localNetwork) addPrimaryValidators(
 		)
 		cancel()
 		if err != nil {
-			return fmt.Errorf("P-Wallet Tx Error %s %w, node ID %s", "IssueAddPermissionlessValidatorTx", err, nodeID)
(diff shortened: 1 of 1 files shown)


local/blockchain.go
```diff
@@ -990,7 +990,7 @@ func (*localNetwork) createBlockchains(
 			defaultPoll,
 		)
 		if err != nil {
-			return fmt.Errorf("failure issuing create blockchain: %w", err)
+			return fmt.Errorf("P-Wallet Tx Error %s %w, blockchainID %s", "Issue Create Blockchain Tx", err, blockchainTxs[i].ID().String())
```
felipemadero, line 993: can you use IssueCreateBlockchainTx instead of "Issue Create Blockchain Tx"? for consistency with the other err msgs
also, probably all IDs in the other err msgs will need a String() call as in this one (eg nodeID.String() etc)

felipemadero (approved):

PR #533, opened 2023-04-18
use macos TMPDIR

felipemadero:


.github/workflows/build-test-release.yml
```diff
@@ -56,14 +56,19 @@ jobs:
       - name: Run e2e tests
         shell: bash
         run: scripts/tests.e2e.sh
+      - name: Set ubuntu tmpdir into actions env
+        if: matrix.os == 'ubuntu-20.04'
```
patrick-ogrady, line 60: Would be good to support `ubuntu-latest` here but

patrick-ogrady: I think you can do a `|| 'ubuntu-22.04'`

patrick-ogrady: or even better, `startsWith(matrix.os, 'ubuntu')`

.github/workflows/build-test-release.yml
```diff
@@ -56,14 +56,19 @@ jobs:
...
+        run: echo "ANR_TMPDIR=$TMPDIR" >> $GITHUB_ENV
       - name: "Upload Artifact"
         if: always()
         uses: actions/upload-artifact@v3
         with:
           name: anr-logs-${{ matrix.os }}
           path: |
-            /tmp/network-runner-root-data*/
-            /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/network-runner-root-data*/
+            /${{ env.ANR_TMPDIR }}/network-runner-root-data*/
```
patrick-ogrady, line 71: I think you'll want to remove the leading `/`

PR #546, opened 2023-04-25
generate prometheus config

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/528

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -885,7 +885,7 @@ func (s *server) PauseNode(ctx context.Context, req *rpcpb.PauseNodeRequest) (*r
 		return nil, err
 	}
 
-	if err := s.network.updateNodeInfo(); err != nil {
+	if err := s.network.UpdateNodeInfo(); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -885,7 +885,7 @@ func (s *server) PauseNode(ctx context.Context, req *rpcpb.PauseNodeRequest) (*r
 		return nil, err
 	}
 
-	if err := s.network.updateNodeInfo(); err != nil {
+	if err := s.network.UpdateNodeInfo(); err != nil {
```
dasconnor, line 888: Was this broken in main?

dasconnor: Are these two different functions?

dasconnor (approved):
Prometheus part looks good, just a little confused why you changed the `updateNodeInfo` function name

server/server.go
felipemadero: Yup the upper one takes a look in network.go. By mistake the lower case one was used. The chance is to be uniform in the usage of the locked version by all server calls (AddNode, RestartNode, etc).

dasconnor: Can you open a ticket to change the name of one of those methods? This seems ripe for future bugs.

felipemadero: https://github.com/ava-labs/avalanche-network-runner/issues/548

PR #559, opened 2023-05-12
control cmd genesis file check add

izzetemredemir:
There was no error message checking if the Genesis file exists I added this for the control command.


izzetemredemir:
Okay. I am going to update the code.

cmd/control/control.go
```diff
@@ -286,6 +287,27 @@ func startFunc(*cobra.Command, []string) error {
 		if err := json.Unmarshal([]byte(blockchainSpecsStr), &blockchainSpecs); err != nil {
 			return err
 		}
+		for _, spec := range blockchainSpecs {
+
+			// Check if spec.Genesis is valid json content
+			if strings.Contains(spec.Genesis, "{") || strings.Contains(spec.Genesis, "}") {
```
felipemadero, line 293: there is a problem related with that the genesis could not necesarily be a json, this is vm dependant. probably we can not do so much with this check. previously ANR was only expecting a file and so an error was propagated from server/server.go, but with the current semantic it seems we just cant do the check and just wait for the vm to return error.

izzetemredemir: Maybe we can consider split parameters with genesis Content and Genesis FilePath.

felipemadero:
Closing for lack of activity. Also path based args will be obsolete soon.

PR #566, opened 2023-05-25
Remove subnet validators

sukantoraymond:


felipemadero (commented):
can you just put the remove val part here, avoiding the add val part?

PR #575, opened 2023-06-12
README: fixup typos and invalid flags

hexfusion:
Updated a few doc bugs I found attempting custom configurations. I believe the  `--plugin-dir` must be defined by `control start`? Regardless it is not a valid flag for `control create-subnets` or `control create-blockchains`.


felipemadero:
> Updated a few doc bugs I found attempting custom configurations. I believe the `--plugin-dir` must be defined by `control start`? Regardless it is not a valid flag for `control create-subnets` or `control create-blockchains`.

Thanks! Yes, it is available as param when starting the network with start or load-snapshot (or when adding/restarting a node). Create blockchains uses the value passed in before.

felipemadero (approved):

PR #611, opened 2023-07-21
Update readme with academy changes

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/610


README.md
```diff
@@ -105,6 +105,10 @@ avalanche-network-runner server \
 
 Note that the above command will run until you stop it with `CTRL + C`. You should run further commands in a separate terminal.
 
+_Note: If you set the environment variables `AVALANCHEGO_EXEC_PATH` and `AVALANCHEGO_PLUGIN_PATH` previously to launching the server,
```
dasconnor, line 108: If the environment variables ... and ... are set, they will be used to start the server if no command line flags are passed.

README.md
```diff
@@ -336,6 +340,28 @@ avalanche-network-runner control create-subnets '[{}, {"participants": ["node1",
...
+
+```bash
+curl -X POST -k http://localhost:8081/v1/control/listsubnets
+
+# or
+avalanche-network-runner control list-subnets
+```
+
+_Note: To create a blockchain, the vm binary for it should be present under `AVALANCHEGO_PLUGIN_DIR`, with a filename equal to the vm id.
+But, ANR receives a vm name as param, and uses it to generate a vm id to look for. So there is a need to get the expected
```
dasconnor, line 353: I'm not sure what the rest of this paragraph means.

felipemadero: ok simplified the parragraph

PR #680, opened 2023-11-16
bump avago + evm at e2e

felipemadero:


scripts/tests.e2e.sh
```diff
@@ -12,9 +12,9 @@ fi
 
 ANR_PATH=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
 
-DEFAULT_VERSION_1=v1.10.10
-DEFAULT_VERSION_2=v1.10.9
-DEFAULT_SUBNET_EVM_VERSION=v0.5.5
+DEFAULT_VERSION_1=v1.10.15
+DEFAULT_VERSION_2=v1.10.14
+DEFAULT_SUBNET_EVM_VERSION=v0.5.9
```
ceyonur, line 17: subnet-evm 0.5.9 won't work with avago v1.10.14. IMO If these different versions need to be run with subnet-evm; we should define two different versions for subnet-evm as well.

ceyonur: https://github.com/ava-labs/subnet-evm#avalanchego-compatibility

felipemadero: DEFAULT_VERSION_ is not used in combination with DEFAULT_SUBNET_EVM_VERSION, it is just used for a node restart test without evm involved

felipemadero: compatibility must exist between DEFAULT_VERSION_1 and DEFAULT_SUBNET_EVM_VERSION only
would you like a change in names, or some comment, to clarify that?

ceyonur: comment would be nice

ceyonur (approved):

PR #687, opened 2023-12-02
allow remote api calls if inside a docker container

felipemadero:
Closes https://github.com/ava-labs/avalanche-network-runner/issues/686

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -142,8 +142,12 @@ func VerifySubnetHasCorrectParticipants(
 	return false
 }
 
-func InsideDockerContainer() (bool, error) {
-	_, err := os.Stat(dockerEnvPath)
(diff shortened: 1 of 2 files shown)


utils/utils.go
```diff
@@ -140,3 +141,14 @@ func VerifySubnetHasCorrectParticipants(
 	}
 	return false
 }
+
+func InsideDockerContainer() (bool, error) {
```
arturrez, line 145: nit: rename to `IsInsideDockerContainer`

utils/utils.go
```diff
@@ -140,3 +141,14 @@ func VerifySubnetHasCorrectParticipants(
...
+	_, err := os.Stat(dockerEnvPath)
```
arturrez, line 146: use utils.FileExists

PR #736, opened 2024-09-25
Avago upgrade

felipemadero:
- re-enable sybil protection for networks of > 1 nodes
- adds upgrade file conf


local/helpers.go
```diff
@@ -36,7 +36,12 @@ func getFreePort() (uint16, error) {
 
 // writeFiles writes the files a node needs on startup.
 // It returns flags used to point to those files.
-func writeFiles(genesis []byte, nodeRootDir string, nodeConfig *node.Config) (map[string]string, error) {
+func writeFiles(
+	genesis []byte,
```
arturrez, line 40: nit: better names `genesisData`, `upgradeData`

local/network.go
```diff
@@ -81,6 +82,9 @@ type localNetwork struct {
 	// This network's genesis file.
```
arturrez, line 82: nit: not the file it's data, so I still think `genesisData` and `upgradeData` are the better names

arturrez (approved):

PR #738, opened 2024-10-11
Add option to generate fresh staker files to Start

felipemadero:
also removes unneeded flags


local/network.go
```diff
@@ -1181,12 +1179,16 @@ func (ln *localNetwork) buildArgs(
 	flags := map[string]string{
 		config.NetworkNameKey: fmt.Sprintf("%d", ln.networkID),
 		config.DataDirKey:     dataDir,
-		config.DBPathKey:      dbDir,
-		config.LogsDirKey:     logsDir,
 		config.PublicIPKey:    publicIP,
 		config.HTTPPortKey:    fmt.Sprintf("%d", apiPort),
 		config.StakingPortKey: fmt.Sprintf("%d", p2pPort),
 	}
+	if dbDir != filepath.Join(dataDir, defaultDBSubdir) {
```
arturrez, line 1186: not clear for me if we need this.we can  just check if dbDir is not ""

local/network.go
```diff
@@ -1181,12 +1179,16 @@ func (ln *localNetwork) buildArgs(
...
+		flags[config.DBPathKey] = dbDir
+	}
+	if logsDir != filepath.Join(dataDir, defaultLogsSubdir) {
```
arturrez, line 1189: same

arturrez (approved):
with 1 small nit

local/network.go
felipemadero: dbDir is not going to be empty from previous logic. the check here is to avoid adding a flag if it is
the default avalanchego value.

local/network.go
felipemadero: same. adding a comment for this.

PR #739, opened 2024-10-11
Varios Fixes

felipemadero:


client/client.go
```diff
@@ -118,13 +118,14 @@ func (c *client) Start(ctx context.Context, execPath string, opts ...OpOption) (
...
-		UpgradeConfigs:  ret.upgradeConfigs,
-		SubnetConfigs:   ret.subnetConfigs,
-		FreshStakingIds: ret.freshStakingIds,
+		NetworkId:              ret.networkID,
+		ExecPath:               execPath,
+		NumNodes:               &ret.numNodes,
+		ChainConfigs:           ret.chainConfigs,
+		UpgradeConfigs:         ret.upgradeConfigs,
+		SubnetConfigs:          ret.subnetConfigs,
+		ZeroIpIfPublicHttpHost: ret.zeroIPIfPublicHttpHost,
```
arturrez, line 127: bad name imho. PublicHttpHost is better

local/blockchain.go
```diff
@@ -1635,7 +1635,7 @@ func (ln *localNetwork) reloadVMPlugins(ctx context.Context) error {
 		if node.paused {
 			continue
 		}
-		uri := fmt.Sprintf("http://%s:%d", node.GetURL(), node.GetAPIPort())
+		uri := fmt.Sprintf("http://%s:%d", node.GetIP(), node.GetAPIPort())
```
arturrez, line 1638: nit: maybe move it to node.GetURI to make it DRY

local/network.go
```diff
@@ -125,6 +128,9 @@ type localNetwork struct {
 	blockchainAliases map[string][]string
 	// wallet private key used. IF nil, genesis ewoq key will be used
 	walletPrivateKey string
+	// nodes always returns 127.0.0.1 as IP
+	// if not set, may return 0.0.0.0 depending on httpHost settings
+	zeroIPIfPublicHttpHost bool
```
arturrez, line 133: bad name

server/network.go
```diff
@@ -817,7 +828,7 @@ func (lc *localNetwork) updateNodeInfo() error {
 
 		lc.nodeInfos[name] = &rpcpb.NodeInfo{
 			Name:               node.GetName(),
-			Uri:                fmt.Sprintf("http://%s:%d", node.GetURL(), node.GetAPIPort()),
+			Uri:                fmt.Sprintf("http://%s:%d", node.GetIP(), node.GetAPIPort()),
```
arturrez, line 831: same: move it for func and make it DRY

local/snapshot.go
```diff
@@ -422,24 +432,39 @@ func (ln *localNetwork) loadSnapshot(
...
+	return RemoveSnapshot(ln.snapshotsDir, snapshotName, snapshotPath)
 }
 
 // Get network snapshots
 func (ln *localNetwork) GetSnapshotNames() ([]string, error) {
 	return GetSnapshotNames(ln.snapshotsDir)
 }
 
-func getSnapshotDir(snapshotsDir string, snapshotName string) string {
+func getSnapshotDir(
```
arturrez, line 447: this is very confusing to me.
the logic behind this func

arturrez: maybe make it simple by making this
```
	if snapshotPath != "" {
		return snapshotPath
	}
	``` a very first statement. followed by
	
	```
	if snapshotsDir == "" {
		snapshotsDir = DefaultSnapshotsDir
	}
	return  filepath.Join(snapshotsDir, **snapshotPrefix+snapshotName)
	```
	too make it much less confusing
	
	pls feel free to ignore this comment

File: local/node_process.go [hunk 6]
```diff
@@ -90,30 +95,46 @@ type nodeProcess struct {
...
 	if err := p.cmd.Start(); err != nil {
 		p.state = status.Stopped
 		close(p.closedOnStop)
+		p.lock.Unlock()
 		return fmt.Errorf("couldn't start process: %w", err)
 	}
 
 	go p.awaitExit()
+	p.lock.Unlock()
+	time.Sleep(startupTime)
```
arturrez, line 130: not sure If I like this approach.
to force it to  sleep for period of time.

we can have a go routine to check on the background if process is started and exit, but not sleep imho

pls feel free to ignore this comment if I didn't read it correctly

local/network.go
```diff
@@ -657,39 +706,88 @@ func (ln *localNetwork) addNode(nodeConfig node.Config) (node.Node, error) {
...
+	if err := json.Unmarshal(processFileBytes, &processContext); err != nil {
+		return node, fmt.Errorf("failed to unmarshal node process context at %s: %w", processFilePath, err)
+	}
+	stakingAddressWords := strings.Split(processContext.StakingAddress, ":")
+	if len(stakingAddressWords) == 0 {
+		return node, fmt.Errorf("unexpected format on staking address %s", processContext.StakingAddress)
+	}
+	p2pPort, err := strconv.ParseUint(stakingAddressWords[len(stakingAddressWords)-1], 10, 16)
+	if err != nil {
+		return node, fmt.Errorf("unexpected format on staking address %s: %w", processContext.StakingAddress, err)
```
arturrez, line 750: p2pPort?

arturrez (commented):
also tests are broken now

local/node_process.go
felipemadero: I want this function to return on error if the process stopped before the given startup time.
So the function is forced to wait that time in the happy path. The only thing I can do with a busy
wait loop is to finish early in the non happy path (the process fails before the startup time). But for that
case I am not so much interested in faster times. So it is a simplest implementation.

felipemadero: Anyway I will reduce this wait to one second as that is the time needed for the failure I am interested in
(ports conflict). And I am adding a busy wait of longer time for the wait for the process context
file, that takes more time in some executions.

local/snapshot.go
felipemadero: making the change

client/client.go
felipemadero: PublicHttpHost is also a bad name. It makes me thinks it is asking the node to be public.
This is not the case. It is asking the node to show his IP as 0.0.0.0 when it is public.
Will call it ZeroIp so the reader can understand it is something unusual.

local/network.go
felipemadero: same response

local/node.go
```diff
@@ -210,11 +214,16 @@ func (node *localNode) GetAPIClient() api.Client {
 }
 
 // See node.Node
-func (node *localNode) GetURL() string {
-	if node.httpHost == "0.0.0.0" || node.httpHost == "." {
+func (node *localNode) GetIP() string {
+	if node.zeroIP && (node.httpHost == "0.0.0.0" || node.httpHost == ".") {
```
arturrez, line 218: should it be || not && ?

arturrez (approved):

PR #742, opened 2024-11-01
use context given by client

felipemadero:


server/server.go
```diff
@@ -1789,3 +1794,23 @@ func applyDefaultPluginDir(userGivenPluginDir string) string {
 	}
 	return os.Getenv(constants.DefaultPluginDirEnvVar)
 }
+
+// Assumes [s.mu] is held.
+func (s *server) getContext(callContext context.Context) (context.Context, context.CancelFunc) {
+	timeout := 3 * time.Minute
```
arturrez, line 1800: nit: should we move timeout into `server struct` as it's a server attribute?
in this case it can be changed from consumer using it (from CLI for example) if needed

arturrez (approved):

server/server.go
felipemadero: I believe it is oks and fairly customizable as it is

